Issue #2, opened 2023-05-02
Add GitHub workflow to run  Zino 2.0 tests


lunkwill42:
Bits to include:

- [ ] Run a test matrix using tox (just like NAV and Argus backend do)
- [x] Run linters to verify that the pre-commit hooks have been used to format code according to Black and PEP-8

PR #5, opened 2023-05-04
Add workflow for linting with flake8 and black

johannaengland:
First part of #2.


lunkwill42 (approved):
While this does the trick, I'm extremely saddened that the GitHub superlinter still doesn't seem to support ruff :(

PR #6, opened 2023-05-04
Add test coverage for zino.snmp

lunkwill42:
This adds initial test coverage for the zino.snmp module.

This was a doozy just to get working.  Although snmpsim has worked perfectly for us in NAV, I had huge problems making it work here, and in a long chain of events, it turned out that:

1. pysnmp is incompatible with the latest pyasn1 release
2. pysnmp doesn't work with Python 3.10 or newer
3. The PySNMP author passed away last year, so the library has been forked in a community effort and an updated version is released on PyPI under a new name: `pysnmplib`.

Once I sorted this out, things fell into place.


lunkwill42:
Seems like the linter workflow applies a different line-length configuration for Black than this project does in `pyproject.toml`, so it fails this PR.  I think @johannaengland is working on fixing the workflow :)

PR #7, opened 2023-05-05
Schedule one extra job to run every minute

stveit:
Instead of scheduling many extra one-time jobs like original zino does, this PR makes it so a job running at a 60 second interval is used instead. While this extra job is running, the reachabilitytask run from the main loop will quit early.

### DIFF
--- a/src/zino/tasks/reachabletask.py
+++ b/src/zino/tasks/reachabletask.py
@@ -10,6 +10,8 @@ _logger = logging.getLogger(__name__)
 
 
 class ReachableTask(Task):
+    EXTRA_JOB_INTERVAL = 60
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._scheduler = get_scheduler()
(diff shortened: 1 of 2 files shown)


src/zino/tasks/reachabletask.py
```diff
@@ -1,67 +1,75 @@
...
+
+    async def _run_extra_job(self):
+        result = await self._get_sysuptime()
+        if result:
+            _logger.debug("Device %s is reachable", self.device.name)
             event = state.events.get(self.device.name, None, EventType.REACHABILITY)
             if event:
                 # TODO update event attributes
+                event.reachability = ReachabilityState.REACHABLE
                 event.add_log(f"{self.device.name} reachable")
```
lunkwill42, line 45: Something of the same sort here: Only log that a device is reachable if its state switched from unreachable to reachable  (although, of course, this may already have that effect, since the extra job is descheduled when the device becomes available again)

lunkwill42 (changes requested):
Looks just about right, but we should the original codebase: Once a device becomes reachable again, should the normal, full collection job be rescheduled to start immediately (to pick up the slack, so to say), or should it continue on its regular schedule?

Anyways, only some small feedback about logging

src/zino/tasks/reachabletask.py
stveit: .. can add similar check here

lunkwill42 (approved):

PR #8, opened 2023-05-05
Make superlinter detect pyproject.toml as config file

johannaengland:
Will remove the last commit after review. That is just to show that the linter is failing when it should fail.


lunkwill42 (approved):
👍

PR #12, opened 2023-05-08
Add explicit path to black and isort config file

johannaengland:
This should fix the problem of super-linter not finding the config file. It needs to be explicitly stated for each linter. 

I tested this with #6.


lunkwill42 (approved):
So I guess explicit *is* still better than implicit :-P

Nice work!

PR #22, opened 2023-06-14
Add simple event registry state persistence

### LINKED ISSUE #21: Persist Zino event state to disk
Event state should be persisted to a disk file on set intervals (and later, on demand).  This state should be loaded from disk whenever the daemon starts up as well.

lunkwill42:
Closes #21.

Depends on #19 being merged first.

This should be expanded upon as soon as we start collecting device state.

### DIFF
--- a/src/zino/state.py
+++ b/src/zino/state.py
@@ -39,3 +39,16 @@ async def dump_state_to_file(filename: str = STATE_FILENAME):
     _log.debug("dumping state to %s", filename)
     with open(filename, "w") as statefile:
         statefile.write(events.json(exclude_none=True, indent=2))
+
+
+def load_state_from_file(filename: str = STATE_FILENAME):
+    """Loads and replaces Zino state from a JSON file dump"""
+    _log.info("Loading saved state from %s", filename)
+    try:
+        loaded_state = Events.parse_file(filename)
+    except FileNotFoundError:
+        _log.error("No state file found (%s), starting from scratch ", filename)
+        return
+    else:
+        global events
+        events = loaded_state
--- a/src/zino/state.py
+++ b/src/zino/state.py
@@ -22,3 +22,18 @@ events = Events()
 
 async def dump_state_to_log():
     _log.debug("Dumping state to log:\n%s", pprint.pformat(events.dict(exclude_none=True)))
+
+
+async def dump_state_to_file(filename: str = "zino-state.json"):
+    """Dumps the event state to a file as JSON.
+
+    Does not dump all state (yet).
+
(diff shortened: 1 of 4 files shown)


src/zino/state.py
```diff
@@ -39,3 +39,16 @@ async def dump_state_to_file(filename: str = STATE_FILENAME):
...
+    """Loads and replaces Zino state from a JSON file dump"""
+    _log.info("Loading saved state from %s", filename)
+    try:
+        loaded_state = Events.parse_file(filename)
+    except FileNotFoundError:
+        _log.error("No state file found (%s), starting from scratch ", filename)
+        return
+    else:
+        global events
+        events = loaded_state
```
stveit, line 54: If this function is unable to complete its purpose (loading state from a file), it should raise an exception (probably raise the FileNotFoundError).

tests/statemodels_test.py
```diff
@@ -17,4 +17,4 @@ def test_add_history_should_set_proper_timestamp(self, fake_event):
 
 @pytest.fixture
 def fake_event():
-    yield Event(id=42, router="example-gw.example.org", event_type=EventType.REACHABILITY, state=EventState.OPEN)
+    yield Event(id=42, router="example-gw.example.org", event_type=ReachabilityEvent, state=EventState.OPEN)
```
johannaengland, line 20:
```suggestion
    yield Event(id=42, router="example-gw.example.org", state=EventState.OPEN)
```

johannaengland (changes requested):

src/zino/state.py
lunkwill42: Only partial agreement: The first time Zino is started, there is no state file to load. Where and how do we handle that case?

tests/statemodels_test.py
lunkwill42: You're commenting on something that was part of #19 here - it is no longer part of this PR...

src/zino/state.py
lunkwill42: I might suggest that Zino should always stop if it doesn't find the state file, but that you can force-start it the very first time it runs, so that it at least generates an empty state file.

The question is whether robustness requirements dictates that it is more important that Zino is running than that it resumes its old state.  Perhaps a requirements discussion with Håvard is in order.

johannaengland (approved):
Manually tested. All works as expected! 

I would suggest adding `zino-state.json` to `.gitignore`.

src/zino/state.py
stveit: If the file not existing has to be handled in this function, then renaming it to `init_state` or something like that would be more fitting

PR #23, opened 2023-06-15
Slightly improve test coverage of Zino binaries and scripts

lunkwill42:


johannaengland (approved):
Looks good and works as expected. Impressive coverage increase of 15%!

The only thing confusing me is why you're testing twice that the zino help screen runs without error? Once in `bin_test` and once in `zino_test`

src/zino/zino.py
```diff
@@ -33,13 +33,13 @@ def init_event_loop(args: argparse.Namespace):
...
 
-def parse_args():
+def parse_args(arguments=None):
     parser = argparse.ArgumentParser(description="Zino is not OpenView")
     parser.add_argument(
         "--polldevs", type=argparse.FileType("r"), metavar="PATH", default="polldevs.cf", help="Path to polldevs.cf"
     )
 
-    args = parser.parse_args()
+    args = parser.parse_args(args=arguments)
```
hmpf, line 42: Nice

tests/conftest.py
```diff
@@ -6,6 +6,27 @@
 from retry import retry
 
 
+@pytest.fixture
+def polldevs_conf(tmp_path):
+    name = tmp_path.joinpath("polldevs.cf")
+    with open(name, "w") as conf:
```
hmpf, line 12: Is this the same format as Zino 1?

hmpf (commented):
Are we using an existing config-format or making our own from scratch?

src/zino/zino.py
hmpf: And it selected the wrong line of course. Nice that we can pass in the args explicitly.

tests/conftest.py
lunkwill42: It is. The first code written for Zino 2 was the `polldevs.cf` parser, with 100% coverage.  And voila, suddenly we needed the test data for other tests as well...

lunkwill42:
> Are we using an existing config-format or making our own from scratch?

We are reusing the existing config file format, @hmpf .  The first code written for Zino 2 was the polldevs.cf parser, with 100% coverage. And voila, suddenly we needed the test data for other tests as well...

lunkwill42:
> The only thing confusing me is why you're testing twice that the zino help screen runs without error? Once in `bin_test` and once in `zino_test`

Indeed. The coverage setup I have in PyCharm doesn't measure coverage of subprocesses, so I was just testing to see whether a more direct test would help coverage numbers there.  Our tox/pytest pipeline measures code coverage also for subprocesses (which are listed as indirect coverage changes on CodeCov), so yep.

There's still more to be tested here - but at least the files have gone from 0% to a higher number :)

PR #27, opened 2023-06-15
Add keyword names to function parameters

johannaengland:
When reading through the code I realized that for someone not familiar with APScheduler (aka me) it was confusing that not all parameters had their keywords accompanying them, which made it harder to understand what was actually happening.


lunkwill42 (approved):
:+1:

PR #33, opened 2023-06-18
Catch error for trying to remove non-existing job from scheduler

johannaengland:
It should of course not happen that we try to remove a non-existing job, but I don't think that should make zino crash completely.

Dependent on #32.


lunkwill42 (approved):
Nice! :)

PR #39, opened 2023-06-21
Establish in-memory device state storage, with enterprise ID as proof-of-concept

### LINKED ISSUE #36: Add to device state if device is juniper or cisco
To be able to do #34 and #37 we need to know if a device is juniper or cisco. This should be saved in the device state.

lunkwill42:
Closes #36 

Instead of storing explicit `is_cisco` or `is_juniper` values as part of the device state variables, this stores the raw enterprise ID value as a device state variable, and instead provides utility properties to detect either of the types we are interested in checking for.

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -72,7 +72,7 @@ class DeviceState(BaseModel):
 
     @property
     def is_juniper(self):
-        return self.enterprise_id == 2626
+        return self.enterprise_id == 2636
 
 
 class DeviceStates(BaseModel):
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -66,6 +66,27 @@ class DeviceState(BaseModel):
     # sawPeer
 
 
+class DeviceStates(BaseModel):
+    """Keeps track of the state of all devices we have polled from"""
+
+    devices: Dict[str, DeviceState] = {}
+
+    def __getitem__(self, item) -> DeviceState:
+        return self.devices[item]
+
+    def __contains__(self, item):
+        return item in self.devices
+
+    def __len__(self):
+        return len(self.devices)
+
+    def get(self, device_name: str) -> DeviceState:
+        """Returns a DeviceState object for device_name, creating a blank state object if none exists"""
+        if device_name not in self:
+            self.devices[device_name] = DeviceState(name=device_name)
+        return self[device_name]
+
+
 class LogEntry(BaseModel):
     """Event log entry attributes. These apply both for 'log' and 'history' lists"""
 
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -37,6 +37,7 @@ class DeviceState(BaseModel):
     """Keep device state"""
 
     name: str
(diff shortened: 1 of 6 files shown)


lunkwill42:
During the writing of this PR, I had the idea that the tasks should not try to access global state, but instead each instance should be fed with some state object from the task running code.

Fewer accesses to global variables is a good thing, and in turn also makes it easier to test the code without doing crazy mocking/patching operations - a test can just inject a plugin with the desired test state.

lunkwill42:
Switching to draft mode, as it still lacks tests

lunkwill42:
Added tests, so switched out of draft mode again

lunkwill42:
Ah, nice. Now the tests are failing on GitHub (running just fine locally). No more time left today to figure out why.

src/zino/statemodels.py
```diff
@@ -65,6 +66,14 @@ class DeviceState(BaseModel):
     # portToLocIfDescr
     # sawPeer
 
+    @property
+    def is_cisco(self):
+        return self.enterprise_id == 9
+
+    @property
+    def is_juniper(self):
+        return self.enterprise_id == 2626
```
johannaengland, line 75:
```suggestion
        return self.enterprise_id == 2636
```

tests/statemodels_test.py
```diff
@@ -15,6 +15,16 @@ def test_add_history_should_set_proper_timestamp(self, fake_event):
         assert isinstance(log.timestamp, datetime.datetime)
 
 
+class TestDeviceState:
+    @pytest.mark.parametrize(
+        "enterprise_id,property_name,expected",
+        [(9, "is_cisco", True), (666, "is_cisco", False), (2626, "is_juniper", True), (666, "is_juniper", False)],
```
johannaengland, line 21:
```suggestion 
        [(9, "is_cisco", True), (666, "is_cisco", False), (2636, "is_juniper", True), (666, "is_juniper", False)],
```

src/zino/tasks/vendor.py
```diff
@@ -0,0 +1,38 @@
...
+_logger = logging.getLogger(__name__)
+ENTERPRISES = (1, 3, 6, 1, 4, 1)
+
+
+class VendorTask(Task):
+    """Fetches and stores state information about a Device's vendor"""
+
+    async def run(self):
+        vendor = await self._get_enterprise_id()
+        _logger.debug("Device enterprise id: %r", vendor)
```
johannaengland, line 17: Shouldn't we also log the device name here? Otherwise this log entry doesn't help much

johannaengland (dismissed):

src/zino/tasks/vendor.py
lunkwill42: Indeed. I'm used to the plugin loggers in nav/ipdevpoll, that automatically include the name of the current job and device in the logs :)

lunkwill42:
I apologise for the potential review inconveniences: I re-based this on the latest master to ensure the test suite will still run.

The necessary changes and other fixups have been squashed.

PR #41, opened 2023-06-23
Convert OID response values to tuples of ints

lunkwill42:
Trying to fetch a sysObjectID using `SNMP.get` will crash, as it doesn't know how to convert a response value type of OID.

This PRs adds a test for that issue, and updates the code to convert such responses to tuples of ints.

### DIFF
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -188,6 +188,8 @@ class SNMP:
             value = int(value)
         elif isinstance(value, univ.OctetString):
             value = str(value)
+        elif isinstance(value, ObjectIdentity):
+            value = value.getOid().asTuple()
         else:
             raise ValueError(f"Could not convert unknown type {type(value)}")
         return MibObject(oid_string, value)
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -36,7 +36,7 @@ def _get_engine():
 @dataclass
 class MibObject:
     oid: str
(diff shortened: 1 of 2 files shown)


src/zino/snmp.py
```diff
@@ -188,6 +188,8 @@ def _object_type_to_mib_object(cls, object_type: ObjectType) -> MibObject:
             value = int(value)
         elif isinstance(value, univ.OctetString):
             value = str(value)
+        elif isinstance(value, ObjectIdentity):
+            value = value.getOid().asTuple()
```
stveit, line 192: typing for `value` in `MibObject` should be updated to reflect this

stveit (commented):
typing for `value` in `MibObject` should be updated to reflect this

johannaengland (approved):

hmpf (approved):

src/zino/snmp.py
lunkwill42: Done!

src/zino/snmp.py
```diff
@@ -2,7 +2,7 @@
 import logging
 import threading
 from dataclasses import dataclass
-from typing import Union
+from typing import Tuple, Union
```
stveit, line 5: you can actually just use `tuple` from 3.9 onwards, saving an import

stveit (approved):

Issue #40, opened 2023-06-22
Update docstring for SNMP commands


stveit:
SNMP command docstrings should be updated to better explain the allowed values.

See https://github.com/Uninett/zino/pull/18#pullrequestreview-1481565405 for relevant discussion

stveit:
closed by #43

PR #43, opened 2023-06-23
Update snmp docstrings

stveit:
Fixex #40


lunkwill42 (changes requested):
Looks good to me, except for a couple of typos :)

lunkwill42:
Added a couple more reviewers, since I'm off for vacation in less than an hour

johannaengland (approved):
Looks good :+1:

PR #44, opened 2023-06-23
Represent oid as tuple in MibObject

stveit:
Relevant for #41. For consistency, oids should be presented the same way.

### DIFF
--- a/src/zino/oid.py
+++ b/src/zino/oid.py
@@ -1,0 +1,63 @@
+"""OID manipulation"""
+
+SEPARATOR = "."
+SEPARATOR_B = b"."
+
+
+class OID(tuple):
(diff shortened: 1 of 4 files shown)


lunkwill42:
@stveit Might I also propose repurposing this class from NAV?

https://github.com/Uninett/nav/blob/cc6dc7ab987d90638119744031736ea1f4e65574/python/nav/oids.py#L24-L80

(Although this is released as GPLv3, the code is entirely owned by us, so we can relicense to Zino's license on the blink of an eye)

lunkwill42 (approved):
Looks just about right!

Since we copied the `OID` class, we might want to copy its tests from NAV too, if any.  Make an issue out of it :)

Issue #42, opened 2023-06-23
Task instances should receive Zino state objects as arguments


lunkwill42:
Global variables should be avoided.  As it happens, Zino currently stores most of its running state in global variables in a global `zino.state` module.  This is also how state is accessed.

As it happens, as part of adding persistent device and IP address state information to Zino, I'm moving most of the existing state variables into a new Pydantic model, so that persistence to disk can easily be initiated by serializing this one single state model.

While single top-level state object still needs to be global in a running Zino process, encapsulating everything in a single top-level object has other advantages too:

Zino tasks should be insulated from the need to access global state from `zino.state`.  I believe that the tasks themselves should rather receive a reference to the top-level `ZinoState` object as part of their `__init__` constructors.  The `run()` implementations can then access state information as `self.state.whatever...`.

This would also make testing of task classes a lot easier, as there is less global state to patch/mock out for tests.  A constructed test state can just be fed to the task class as it is being instantiated.

Implementation of this can likely not begin until the `ZinoState` PR is complete, though.

_Idea originally posted by @lunkwill42 in https://github.com/Uninett/zino/issues/39#issuecomment-1600850290_

PR #45, opened 2023-06-23
Encapsulate global state into a ZinoState class

### LINKED ISSUE #35: Persist Zino device state to disk
Device state should be persisted to a disk file on set intervals (and later, on demand). This state should be loaded from disk whenever the daemon starts up as well. Same as #21, just with device state instead of event state.

lunkwill42:
This primarily changes how global state is handled, by encapsulating all state that needs to be persistent between Zino runs into a new ZinoState class (a Pydantic model).

The encapsulation includes the old `zino.state.devices` object, so that device state is now also persisted to disk.  I.e. : This closes #35.

### DIFF
--- a/tests/tasks/test_vendor.py
+++ b/tests/tasks/test_vendor.py
@@ -3,15 +3,15 @@ from unittest.mock import patch
 import pytest
 
 from zino.config.models import PollDevice
-from zino.statemodels import DeviceStates
+from zino.state import ZinoState
 from zino.tasks.vendor import VendorTask
 
 
 class TestVendorTask:
-    @patch("zino.state.devices", DeviceStates())
+    @patch("zino.state.state", ZinoState())
     @pytest.mark.asyncio
     async def test_run_should_set_enterprise_id(self, snmpsim, snmp_test_port):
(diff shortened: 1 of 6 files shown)


lunkwill42:
While all the tests are running successfully, I'm not necessarily convinced that this updates everything that needs to be updated, and since I'm off for vacation, I want to leave this draft in the capable hands of the rest of our team.

These changes are necessary to move on with several of our issues, including #42 - so if something is missing here, please try to fix it and get it merged while I'm away.

tests/tasks/test_vendor.py
```diff
@@ -20,10 +20,10 @@ async def test_run_should_set_enterprise_id(self, snmpsim, snmp_test_port):
         # The "public" test fixture is from an HP switch
         assert state.devices[device.name].enterprise_id == 11
 
-    @patch("zino.state.devices", DeviceStates())
+    @patch("zino.state.state", ZinoState())
```
lunkwill42, line 23: This is the type of silly patching that will become obsolete if we implement #42

johannaengland (approved):

PR #47, opened 2023-06-26
Update reachability event attribute for new job

johannaengland:
When a reachability task job is run and the job is marked an not reachable, an event is created with the reachability attribute `no-response` and another task is scheduled. That state is saved to file.
Now I stopped zino, to update the community of that device and started zino again. Now the reachability task job is run, but the event reachability attribute was not updated. This PR makes sure that it is updated.

And a JobLookupError is caught when removing the extra job, just in case, as we do it in `deschedule_deleted_devices`.

### DIFF
--- a/src/zino/tasks/reachabletask.py
+++ b/src/zino/tasks/reachabletask.py
@@ -37,13 +37,13 @@ class ReachableTask(Task):
             self._schedule_extra_job()
         else:
             _logger.debug("Device %s is reachable", self.device.name)
-            self._update_reachability_event()
+            self._update_reachability_event_as_reachable()
 
     async def _run_extra_job(self):
         result = await self._get_sysuptime()
         if result:
             _logger.debug("Device %s is reachable", self.device.name)
-            self._update_reachability_event()
+            self._update_reachability_event_as_reachable()
             self._deschedule_extra_job()
 
     async def _get_sysuptime(self):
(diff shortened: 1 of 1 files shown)


src/zino/tasks/reachabletask.py
```diff
@@ -66,7 +70,10 @@ def _schedule_extra_job(self):
 
     def _deschedule_extra_job(self):
         name = self._get_extra_job_name()
-        self._scheduler.remove_job(job_id=name)
+        try:
+            self._scheduler.remove_job(job_id=name)
+        except JobLookupError:
+            pass
```
stveit, line 76: 👍

src/zino/tasks/reachabletask.py
```diff
@@ -35,25 +37,27 @@ async def run(self):
...
     async def _get_sysuptime(self):
         snmp = SNMP(self.device)
         result = await snmp.get("SNMPv2-MIB", "sysUpTime", 0)
         return result
 
+    def _update_reachability_event(self):
+        event = state.events.get(self.device.name, None, ReachabilityEvent)
+        if event and event.reachability != ReachabilityState.REACHABLE:
+            event.reachability = ReachabilityState.REACHABLE
+            event.add_log(f"{self.device.name} reachable")
```
stveit, line 58: This only updates the event to a `REACHABLE` state, so the function name should reflect that. updating the event could mean setting it to `NORESPONSE`, changing state to `OPEN` etc

stveit (changes requested):

stveit (approved):

PR #48, opened 2023-06-27
Add task to fetch and log juniper alarm count

### LINKED ISSUE #34: Add task for juniper alarms
For all juniper devices zino should use SNMP to get `jnxYellowAlarmCount` and `jnxRedAlarmCount` and if the alarm count has changed copy it to the device state and to an alarm event.

johannaengland:
Closes #34.


tests/tasks/test_juniperalarmtask.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    task = JuniperAlarmTask(device=device)
+
+    assert (await task.run()) is None
+
+
+@patch("zino.state.devices", DeviceStates())
+@pytest.mark.asyncio
+async def test_juniper_alarm_does_nothing_for_non_juniper_device(snmpsim, snmp_test_port):
+    from zino import state
```
stveit, line 27: after #45, this should now be 
```suggestion
@patch("zino.state.state", ZinoState())
@pytest.mark.asyncio
async def test_juniper_alarm_does_nothing_for_non_juniper_device(snmpsim, snmp_test_port):
    from zino.state import state
```
 or something like that

src/zino/tasks/juniperalarmtask.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        if red_created:
+            red_alarm_event.state = EventState.OPEN
+            red_alarm_event.add_history("Change state to Open")
+        if red_alarm_event.alarm_count != red_alarm_count:
+            old_red_alarm_count = red_alarm_event.alarm_count
+            red_alarm_event.alarm_count = red_alarm_count
+            red_alarm_event.add_log(
+                f"{self.device.name} red alarms went from {old_red_alarm_count} to {red_alarm_count}"
+            )
```
stveit, line 67: this function is pretty big, would prob be nice to split into one function handling red alarms and another handling yellow  or something along those lines

tests/tasks/test_juniperalarmtask.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.mark.asyncio
+async def test_juniper_alarm_updates_alarm_events(snmpsim, snmp_test_port):
+    from zino import state
+
+    device = PollDevice(
+        name="buick.lab.example.org",
+        address="127.0.0.1",
+        community="juniper-alarm",
+        port=snmp_test_port,
+    )
```
stveit, line 128: this could be defined as a fixture so you dont have to define the same thing in every test

src/zino/tasks/juniperalarmtask.py
```diff
@@ -0,0 +1,60 @@
...
+
+    async def _get_juniper_alarms(self):
+        snmp = SNMP(self.device)
+        yellow_alarm_count = await snmp.get("JUNIPER-ALARM-MIB", "jnxYellowAlarmCount", 0)
+        red_alarm_count = await snmp.get("JUNIPER-ALARM-MIB", "jnxRedAlarmCount", 0)
+        if yellow_alarm_count:
+            yellow_alarm_count = yellow_alarm_count.value
+        if red_alarm_count:
+            red_alarm_count = red_alarm_count.value
+        return yellow_alarm_count, red_alarm_count
```
stveit, line 46: Related to previous comment, should probably handle here if `yellow_alarm_count` is None or `yellow_alarm_count.value` is not an `int`etc, and raise exception if something is wrong

tests/tasks/test_juniperalarmtask.py
```diff
@@ -0,0 +1,113 @@
...
+
+from zino.config.models import PollDevice
+from zino.events import AlarmEvent
+from zino.state import ZinoState
+from zino.tasks.juniperalarmtask import JuniperAlarmTask
+
+
+class TestJuniperalarmTask:
+    @pytest.mark.asyncio
+    async def test_juniper_alarm_runs_without_errors(self, snmpsim, juniper_alarm_task):
```
stveit, line 11: Comment to all test names: Since the class name states that these are test for the juniper alarm task, the test names dont need to include too much context, so `test_task_runs_without_errors` or something like that should be enough. No need to state in class name and test names that its Juniper Alarm task specifically

stveit (changes requested):

src/zino/tasks/__init__.py
```diff
@@ -5,7 +5,8 @@ async def run_all_tasks(device, state):
 
 
 def get_registered_tasks():
+    from zino.tasks.juniperalarmtask import JuniperAlarmTask
     from zino.tasks.reachabletask import ReachableTask
     from zino.tasks.vendor import VendorTask
 
-    return [ReachableTask, VendorTask]
+    return [JuniperAlarmTask, ReachableTask, VendorTask]
```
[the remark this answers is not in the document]
johannaengland: I was thinking about having this alphabetical, which is why I put it first

src/zino/tasks/juniperalarmtask.py
```diff
@@ -0,0 +1,60 @@
...
+        yellow_alarm_count, red_alarm_count = await self._get_juniper_alarms()
+
+        if type(yellow_alarm_count) is not int or type(red_alarm_count) is not int:
+            _logger.error(
+                "Device %s returns an alarm count that is not of type int, yellow alarm: %r, red alarm count: %r",
+                self.device.name,
+                yellow_alarm_count,
+                red_alarm_count,
+            )
+            return
```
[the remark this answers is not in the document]
johannaengland: The question is - should we actually raise an error if the alarm count is not an int?

src/zino/tasks/__init__.py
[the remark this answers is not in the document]
stveit: Aah I see. Well, for now it doesnt really matter so whatever order you prefer is fine.

In the future tho the order might matter, if we want to make it so other tasks are cancelled if ReachableTask fails (no reason to fetch alarms from a device thats not reachable etc)

src/zino/tasks/juniperalarmtask.py
```diff
@@ -0,0 +1,64 @@
...
+            red_alarm_count = red_alarm_count.value
+
+        if type(yellow_alarm_count) is not int or type(red_alarm_count) is not int:
+            _logger.error(
+                "Device %s returns an alarm count that is not of type int, yellow alarm: %r, red alarm count: %r",
+                self.device.name,
+                yellow_alarm_count,
+                red_alarm_count,
+            )
+            raise TypeError
```
stveit, line 48: Maybe include in the log the type of the alarm counts instead of the values themselves, since the point of the error message is that the types are wrong.

stveit (changes requested):

src/zino/tasks/__init__.py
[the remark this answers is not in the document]
johannaengland: True, I haven't thought about that, good call!

src/zino/tasks/juniperalarmtask.py
```diff
@@ -0,0 +1,64 @@
...
+        yellow_alarm_count = await snmp.get("JUNIPER-ALARM-MIB", "jnxYellowAlarmCount", 0)
+        red_alarm_count = await snmp.get("JUNIPER-ALARM-MIB", "jnxRedAlarmCount", 0)
+        if yellow_alarm_count:
+            yellow_alarm_count = yellow_alarm_count.value
+        if red_alarm_count:
+            red_alarm_count = red_alarm_count.value
+
+        if type(yellow_alarm_count) is not int or type(red_alarm_count) is not int:
+            _logger.error(
+                "Device %s returns alarm count not of type int, type yellow alarm: %s, type red alarm count: %s",
```
stveit, line 43: final comment i think: for yellow it says `yellow alarm` and for red it says `red alarm count`

stveit (approved):

PR #58, opened 2023-07-05
Support pydantic v2

### LINKED ISSUE #56: Update code to work with pydantic v2
Pydantic v2 is now out, and it would be great to make use of the new features. Version 2 breaks the current code base (#54), so alot of the code using pydantic needs to be updated if we are to support v2.

stveit:
fixes #56 

This updates pydantic to v2 and fixes deprecated stuff that caused errors. 
This PR does not update things that still work but could be updated using new pydantic features.

Separate issues and PRs should be made for making use of new features to improve the code base.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 ]
 dependencies = [
     "apscheduler",
-    "pydantic<2",
+    "pydantic>=2",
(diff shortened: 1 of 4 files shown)


johannaengland (approved):
Please have a look at my last commit, then I'm happy with this

PR #68, opened 2023-08-17
Only post juniper alarm events on changed state

### LINKED ISSUE #67: Juniper alarms task creates events no matter whether there has been any state change
### What did I observe
The `JuniperAlarmTask` was `imlemented` in #48, but it appears to have a glaring mistake:

https://github.com/Uninett/zino/blob/7f3e1f856e68567e5b683c4b92f925af972c5339/src/zino/tasks/juniperalarmtask.py#L29-L30

Looking at the code surrounding these two lines, there isn't actually any part of this code that compares the collected alert counts to any existing zino state for this device.

The immediate effect here is that when Zino starts for the first time, two new events will be opened for every registered Juniper router in `polldevs.cf`, and these will all log something like `{router} {color} alarms went from None to {count}` - where, for most routers, `{count}` is zero.

## Expected behavior

The first time a router is added to `polldevs.cf`, Zino doesn't have any prior state information about the router's alert counts.  They should be assumed to be 0 on the first run.

Only when the polled alert count *differs* from the previously collected count value should an event be created (or an existing event be updated with more log entries).

Starting new devices with the assumption of 0-value counts will ensure that new events are created only for routers that actually have active alerts.

johannaengland:
Closes #67.

Also initializes device alarm state with zero on first run, amends existing tests to accommodate that change and adds new tests.


src/zino/tasks/juniperalarmtask.py
```diff
@@ -21,13 +21,19 @@ async def run(self):
...
 
-        device_state.alarms = {
-            "yellow": yellow_alarm_count,
-            "red": red_alarm_count,
-        }
+        if not device_state.alarms:
+            device_state.alarms = {
+                "yellow": 0,
+                "red": 0,
+            }
```
stveit, line 28: Would it maybe make more sense to set default values yellow=0 and red=0 in the DeviceState model itself rather than doing it here?

src/zino/tasks/juniperalarmtask.py
```diff
@@ -40,10 +46,14 @@ async def _get_juniper_alarms(self):
...
             _logger.error(
-                "Device %s returns alarm count not of type int, type yellow alarm count: %s, type red alarm count: %s",
+                "Device %s returns alarm count not of type int. "
+                "Yellow alarm count: type %s, value %s. "
+                "Red alarm count: type %s, value %s.",
                 self.device.name,
                 type(yellow_alarm_count),
+                yellow_alarm_count,
                 type(red_alarm_count),
+                red_alarm_count,
```
stveit, line 56: Is the alarm count value relevant in an error message regarding the typing itself?

johannaengland: Yes, since @lunkwill42 explicitly stated that he is interested in that information

src/zino/tasks/juniperalarmtask.py
johannaengland: I think since we don't have any other alarm counts (and I can't think of any to follow soon) it is more useful to set it here to not have so much extra unnecessary state information

src/zino/tasks/juniperalarmtask.py
lunkwill42: It could easily be moved to a separate debug statement (which might be more befitting).  But one thing I would like to suggest here is to use `%r` for the values, not `%s`, as that will usually be a more meaningful representation of an unexpected value.

src/zino/tasks/juniperalarmtask.py
lunkwill42: We decided against putting them in as default values, for the time being, namely because of their Juniper-specific nature (our research network is mostly Juniper routers these days, though).

src/zino/tasks/juniperalarmtask.py
```diff
@@ -40,10 +46,14 @@ async def _get_juniper_alarms(self):
 
         if type(yellow_alarm_count) is not int or type(red_alarm_count) is not int:
             _logger.error(
-                "Device %s returns alarm count not of type int, type yellow alarm count: %s, type red alarm count: %s",
+                "Device %s returns alarm count not of type int. "
+                "Yellow alarm count: type %s, value %s. "
+                "Red alarm count: type %s, value %s.",
```
lunkwill42, line 51:
```suggestion
                "Yellow alarm count: type %s, value %r. "
                "Red alarm count: type %s, value %r.",
```

Also, given @stveit's comments, we might consider moving the nitty-gritty details to a separate debug log statement...

tests/tasks/test_juniperalarmtask.py
```diff
@@ -77,7 +77,26 @@ async def test_task_creates_alarm_events(self, snmpsim, juniper_alarm_task):
 
         assert yellow_event
         assert red_event
-        assert yellow_event.alarm_count == 0
+        assert yellow_event.alarm_count == 1
+        assert red_event.alarm_count == 2
+
+    @pytest.mark.asyncio
+    async def test_task_creates_one_alarm_event_on_one_count_changed(self, snmpsim, juniper_alarm_task):
```
lunkwill42, line 84: A hint: If you do not actually need to reference the snmpsim fixture content directly from your test, you don't need to add it to your test signature: `juniper_alarm_task` already depends on the `snmpsim` fixture, so it will already be present.  That's the beauty of fixtures being able to depend on other fixtures.

lunkwill42 (changes requested):
Looks like it will do the trick, just the usual nitpicks from me :-)

lunkwill42 (approved):
Thank you!

Issue #37, opened 2023-06-21
Make a link state monitor task


johannaengland:
The link state monitor basically:

1. Collects the complete list of all ports on a device, including their link state
2. Detects link state changes since the last collection and makes new events for changed states

### Implementation breakdown

- [x] #69
- [x] #70
- [x] #75
- [x] #77
- [x] #79

lunkwill42:
Noting that the original Zino code collects both `locIfDescr` and `ifAlias` IF-MIB values if available. However, as pointed out by Håvard in a discussion earlier this year, `locIfDescr` has been deprecated for a number of years (remember, Tcl-Zino started in the 90s), and has been supplanted by `ifAlias`.

I.e. we shouldn't need to care about `locIfDescr` at all in Zino 2.

lunkwill42:
Exploratory coding for this issue reveals that more changes are needed to the base SNMP adapter in `zino.snmp.SNMP`.  #69 is one of these.

Also, none of the current walk implementations support more than one walk variable, while the underlying protocol supports asking for more than one variable per request (whether it be GET-NEXT or GET-BULK).  

The link state monitor needs to get a bunch of attributes (i.e. columns) per interface, and the best way to do this SNMP-wise is to do what Zino 1 calls a sparse walk.  To avoid implementing a sparse walk as a bunch of daisy-chained walks of individual columns, a new getbulk-implementation is needed in `zino.snmp.SNMP`.
,

lunkwill42:
I've removed the flapping task from this issue, as the flapping detection algorithm is only relevant to SNMP trap processing, which isn't implemented yet. This issue can therefore be closed, as the active link state monitor is more or less complete.

PR #70, opened 2023-08-22
Implement a "sparsewalk" operation for zino.snmp.SNMP

lunkwill42:
As identified by #37, the SNMP adapter library needs an "sparsewalk" SNMP operation, akin to the one used in the Tcl version of Zino.


lunkwill42:
Ah, now we have the lovely case of tests failing on GitHub, while tox runs perfectly locally :-P I think I'll call it a day now...

johannaengland (approved):
As soon as `IF-MIB` is actually added this can be merged

lunkwill42:
So, tests were failing mainly because IF-MIB could not be resolved in the CI environment, while it worked on my system.  This is likely mainly due to the fact that PySNMP caches all the Python modules it builds from ASN.1 sources to `~/.pysnmp`, and I did have `IF-MIB.py` there.

PySNMP's mechanism for adding ASN.1 sources, however, seems pretty lame compared to how you can add more pre-compiled Python MIB sources.  The latter can be added to a MibBuilder object and using in an SNMP engine.  The former can only be added to individual `ObjectIdentity` instances by calling its `addAsn1MibSource()` method.

I hope this doesn't turn into a nightmare of test order dependencies, though.  I suggest that we add a resolver test for each new MIB that we compile and include, so the test suite can give some guarantees that we can actually resolve the MIB data we need.

PR #76, opened 2023-10-04
Raise and handle exceptions in SNMP module

### LINKED ISSUE #73: zino.snmp.SNMP needs improved error handling and logging
The current state of error handling and error logging in `zino.snmp.SNMP` leaves a lot to be desired.

There are currently three types of errors that are handled (or mishandled):

### MIB lookup exceptions

These are caught, logged and then ignored, leading to the query functions returning just empty responses.  The logging bit is fine, but should really include information about which operation was called (get/getnext/bulk/whatever).

However, the lookup error should either be re-raised, or wrapped in a Zino-level exception for the calling code to handle.

### Error responses from SNMP agents.

If a problematic SNMP query is received by an agent, it will typically send an error response.  These responses indicate what type of error is reported, and, depending on the error, may also indicate which of the queried in the query packet was to blame (typically, an error index value, where the index is to the list of queries variables).

`SNMP` mostly uses the `_handle_errors` function for this.  This function has several issues: It is not very detailed about what the actual error condition is, or which variable it is located in.  It also doesn't say anything about which type of request (get/bulk/so-on) was sent that caused the error.

Most importantly: These errors are also silently ignored after being logged, resulting in an empty response to the calling code.

These types of errors should also result in some exception that the calling code needs to handle.

There's also this strange line:
https://github.com/Uninett/zino/blob/7e8fa777229b9a48f326e585b4d5704e34b8c5ff/src/zino/snmp.py#L107

We don't run mypy on our codebase yet, but this is flagged in my IDE because `error_status` is expected to be a `str`, according to the type annotations (and according to PySNMP docs), so calling `prettyPrint()` on it shouldn't work.

### Timeouts

Actually, I'm not sure how these are handled.  I would assume that the `ReachableTask` would receive and handle a timeout error as a non-reachable event, but it rather seems that the `SNMP.get` method returns an empty response on a timeout.  IMO, when a request times out, a `TimeoutError` should be raised, and the caller should be prepared to handle that.

stveit:
Fixes #73 

This PR makes snmp functions raise exceptions if something goes wrong.
A hierarchy of exceptions for snmp related errors is added for this purpose.

This will cause tasks to fail unless its natural for the task to handle certain exceptions.
For example, it will be natural for ReachableTask to handle TimeoutError, since detecting 
a device not answering is the purpose of the task. VendorTask on the other hand can not function if
a timeout occurs, so it will be natural for this task to let the exception raise further up the chain.

Currently this will cause zino to crash, so a new issue will be needed for handling tasks failing.

### DIFF
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -383,7 +383,7 @@ class SNMP:
 
     @property
     def udp_transport_target(self) -> UdpTransportTarget:
-        return UdpTransportTarget((str(self.device.address), self.device.port))
+        return UdpTransportTarget((str(self.device.address), self.device.port), timeout=self.device.timeout)
 
 
 def _convert_varbind(ident: ObjectIdentity, value: ObjectType) -> SNMPVarBind:
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -18,6 +18,7 @@ from pysnmp.hlapi.asyncio import (
     getCmd,
     nextCmd,
 )
+from pysnmp.proto.errind import RequestTimedOut
 from pysnmp.smi import builder, view
 from pysnmp.smi.error import MibNotFoundError as PysnmpMibNotFoundError
 
@@ -100,26 +101,22 @@ class SNMP:
             )
         except PysnmpMibNotFoundError as error:
             raise MibNotFoundError(error)
-        if self._handle_errors(error_indication, error_status, error_index, query):
-            return
+        self._raise_errors(error_indication, error_status, error_index, query)
         for var_bind in var_binds:
             return self._object_type_to_mib_object(var_bind)
 
-    def _handle_errors(self, error_indication: str, error_status: str, error_index: int, *query: ObjectType) -> bool:
-        """Returns True if error occurred"""
+    def _raise_errors(self, error_indication: str, error_status: str, error_index: int, *query: ObjectType) -> bool:
+        """Raises a relevant exception if an error has occurred"""
+        # Local errors (timeout, config errors etc)
         if error_indication:
-            _log.error("%s: %s", self.device.name, error_indication)
-            return True
+            if isinstance(error_indication, RequestTimedOut):
+                raise TimeoutError(str(error_indication))
 
+        # Remote errors from SNMP entity.
+        # if nonzero error_status, error_index point will point to the
+        # variable-binding in query that caused the error.
         if error_status:
-            _log.error(
-                "%s: %s at %s",
-                self.device.name,
-                error_status.prettyPrint(),
-                error_index and query[int(error_index) - 1][0] or "?",
-            )
-            return True
-        return False
+            pass
 
     async def getnext(self, *oid: str) -> Union[MibObject, None]:
         """SNMP-GETNEXTs the given oid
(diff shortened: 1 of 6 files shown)


src/zino/snmp.py
```diff
@@ -161,21 +192,17 @@ async def walk(self, *oid: str) -> list[MibObject]:
...
             current_object = await self._getnext(current_object)
-            if not current_object or not original_oid.is_a_prefix_of(str(current_object[0])):
+            if not original_oid.is_a_prefix_of(str(current_object[0])):
                 break
             mib_object = self._object_type_to_mib_object(current_object)
             results.append(mib_object)
         return results
 
-    async def getbulk(self, *oid: str, max_repetitions: int = 1) -> list[MibObject]:
+    async def getbulk(self, *oid: str, max_repetitions: int = 10) -> list[MibObject]:
```
lunkwill42, line 205: Why was this 1 to begin with, and why did it change to 10 in a PR focused on error handling?

src/zino/snmp.py
```diff
@@ -362,7 +378,9 @@ def community_data(self) -> CommunityData:
 
     @property
     def udp_transport_target(self) -> UdpTransportTarget:
-        return UdpTransportTarget((str(self.device.address), self.device.port))
+        return UdpTransportTarget(
+            (str(self.device.address), self.device.port), timeout=self.device.timeout, retries=self.retries
```
lunkwill42, line 382: Again, there should already be a `self.device.retries` :)

src/zino/tasks/reachabletask.py
```diff
@@ -21,8 +21,9 @@ async def run(self):
         """Checks if device is reachable. Schedules extra reachability checks if not."""
         if self._extra_job_is_running():
             return
-        result = await self._get_sysuptime()
-        if not result:
+        try:
+            await self._get_sysuptime()
+        except TimeoutError:
```
lunkwill42, line 26: A note for future reference (not to be fixed by this PR): Depending on the original Zino implementation, this function should probably record the device boot time to its state.  I've added a utility function to `DeviceState` for this in one of the PRs I'm currently working on, as the original link state checker did this.

lunkwill42 (changes requested):
This looks like a very nice improvement on the SNMP adapter code!

On some level, I kind of like the way you've grouped the various Exception-related tests in classes, but I also find it makes it somewhat hard to reason about which tests that exercise a function I'm looking at. Maybe not a big problem with modern IDEs (also, since I've taken to look at *mutmut*, I've had to switch on dynamic contexts for coverage measurements, so that mutmut can trace which tests cover which code lines)

src/zino/snmp.py
```diff
@@ -59,23 +61,57 @@ class Identifier(NamedTuple):
...
+    pass
+
+
 class SNMP:
     """Represents an SNMP management session for a single device"""
 
     NON_REPEATERS = 0
 
-    def __init__(self, device: PollDevice):
+    def __init__(self, device: PollDevice, retries=5):
```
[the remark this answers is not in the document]
stveit: completely missed that retries was in polldevice

src/zino/snmp.py
lunkwill42: In order to avoid confusion: For bulk requests, I think generally a default value higher than 1 makes a lot of sense (the whole point of the bulk request is normally to fetch more than just 1 value!).

So I think we *should* increase the number.  I just don't think it's relevant for this specific PR.

lunkwill42:
> I have a pretty strong suspicion all the return values from the various SNMP commands (getCmd, bulkCmd) etc, all have misleading docstrings. I am pretty sure `error_status` is returned as an `int` and not a `str`. I also think `error_indication` is `None` if there are no errors.

Yeah, the library is confusing enough at certain points.  It could do with some proper type annotations, at least.  I did read somewhere in those docs just yesterday that `error_indication` is a string, when it is in fact an Exception instance.

> I am leaving any more fixes for this to a separate issue #82

Great :)

lunkwill42 (approved):
:+1:

PR #78, opened 2023-10-04
Use sysUpTime to calculate and log interface state change times

### LINKED ISSUE #77: Use `ifLastChange` to calculate link state change times

lunkwill42:
Fixes #77

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -46,9 +46,9 @@ class DeviceState(BaseModel):
     boot_time: Optional[int] = None
     ports: Dict[int, Port] = {}
     alarms: Optional[Dict[AlarmType, int]] = None
+    boot_time: Optional[datetime.datetime] = None
 
     # This is the remaining set of potential device attributes stored in device state by the original Zino code:
-    # BootTime
     # EventId
     # JNXalarms
     # RunsOn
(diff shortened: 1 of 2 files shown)


src/zino/tasks/linkstatetask.py
```diff
@@ -39,10 +40,13 @@ class LinkStateTask(Task):
     case: Descriptions are mandated for both physical ports and their sub-units.
     """
 
+    sysuptime: Optional[int] = 0
```
hmpf, line 43: What's the difference between sysuptime set to 0 and None?

hmpf (commented):
See inline question.

src/zino/tasks/linkstatetask.py
lunkwill42: `None` would usually mean that we don't know the `sysUpTime` value, while `0` would mean that the device booted *just now*.

However, the `Optional` here seems to be redundant. Within this plugin, the value isn't optional.  The original link state code basically defaults to using 0 as the base for calculating timestamps if the sysuptime value couldn't be fetched.

src/zino/statemodels.py
```diff
@@ -80,6 +80,13 @@ def is_cisco(self):
     def is_juniper(self):
         return self.enterprise_id == 2636
 
+    def set_boot_time_from_uptime(self, uptime: int):
+        """Calculates and sets the device boot time from a current uptime value.
+
+        :param uptime: An uptime value in 100ths of a second
+        """
+        self.boot_time = datetime.datetime.now() - datetime.timedelta(seconds=uptime / 100)
```
lunkwill42, line 88: Actually, I would change this slightly, now that I see it again in hindsight.

We already have a utility function to ensure we generate timestamps with timezone info: `zino.time.now()`

hmpf (approved):

PR #80, opened 2023-10-05
Implement getnext capable of multi-var query

lunkwill42:
This adds an alternative getnext implementation (getnext2), which is capable of dispatching multi-varbind queries, and capable of translating the response varbinds to symbolic names.

PR #83, opened 2023-10-06
Schedule imminent link state verification on link state changes

lunkwill42:
This adds the necessary functionality to the linkstate monitor to replicate what the original Zino code does:

Every time a link state change is detected for a port, a re-verification for that single port is scheduled to run in two minutes.

This depends on #71, #78 AND #80 before it can be merged, so keeping in Draft mode for now.


src/zino/tasks/linkstatetask.py
```diff
@@ -51,6 +57,20 @@ async def run(self):
...
+        poll_list = [("IF-MIB", column, str(ifindex - 1)) for column in BASE_POLL_LIST]
+        result = await snmp.getnext2(*poll_list)
+        self.sysuptime = await self._get_uptime(snmp)
+        _logger.debug("poll_single_interface %s result: %r", self.device.name, result)
+
+        assert all(ident.index == OID(f".{ifindex}") for ident, value in result)
+        row = {ident.object: value for ident, value in result}
+
+        self._update_single_interface(row)
+        return row
```
stveit, line 72: Is this actually supposed to return something? I cant see the return value being used anywhere

src/zino/tasks/linkstatetask.py
```diff
@@ -51,6 +57,20 @@ async def run(self):
 
         self._update_interfaces(attrs)
 
+    async def poll_single_interface(self, ifindex: int):
+        """Polls and updates a single interface"""
+        snmp = SNMP(self.device)
+        poll_list = [("IF-MIB", column, str(ifindex - 1)) for column in BASE_POLL_LIST]
+        result = await snmp.getnext2(*poll_list)
```
stveit, line 64: would it not be better to use a get operation instead of a getnext? I assume getnext2 is used here since the current  get function does not support multiple values. I think it would be better to either accept the slight inefficiency and do multiple get requests, or make a get2. Correct me if im wrong on my assumptions.

stveit (changes requested):

src/zino/tasks/linkstatetask.py
lunkwill42: Actually, it's re-implemented precisely in the same way Zino 1 does it - and yes, one of the reasons for that is that we get multiple values in a single request.

Go right ahead and make a `get2`, if you'd like, but it's out of scope for this PR.  Switching to individual get requests would require 6 pairs of request/reply packets rather than one pair, and these things begin to add up when there are a lot of routers in the mix.

src/zino/tasks/linkstatetask.py
lunkwill42: Naw. I might have kept it there for testing and forgotten to remove it :)

stveit (approved):

Issue #65, opened 2023-08-15
Make a BFD monitor task


lunkwill42:
Subtasks:

- [x] #86 
- [x] #126

lunkwill42:
Håvard has indicated dissatisfaction with the current BFD monitor in Zino 1.  He should be consulted to design a better implementation for Zino 2.0

stveit:
A juniper implementation has been made (#86), Cisco is in the works

lunkwill42:
Both subtasks have been closed, also closing this tracking issue.

PR #86, opened 2023-10-11
Add BFD monitor task (Juniper)

stveit:
#65 

Reduced the scope of this PR from supporting Cisco and Juniper, to just Juniper. Cisco can come after.

Gets BFD state from device.

 Juniper maps BFD sessions to the relevant interface using the IfDescr value. The interface index is not known, and Ports in Zino require the index to be registered. If BFDTask was to create Ports as it needed them, it would have to go through the ifTable in order to map ifdescr to index.

The link state monitor task #71 populates Ports in Zino and saves the IfDescr value to port.ifdescr. To avoid collecting interface data twice, the BFD monitor task will rely on Ports already being populated, and map a BFD session to the correct Port using the ifdescr value. This means that if a BFD session exist for an interface that is, for some reason, not already registered as a Port in Zino, it will be ignored.

Also not sure if all the added MIBS are needed or if some of them already exist as part of pysnmp.

Event creation is simplified to: State changes, event is made. More advanced logic for event creation should be discussed at some point


src/zino/statemodels.py
```diff
@@ -179,9 +179,9 @@ class BGPEvent(Event):
 
 class BFDEvent(Event):
     type: Literal["bfd"] = "bfd"
-    bfdix: Optional[int] = None
-    bfddiscr: Optional[int] = None
-    bfdaddr: Optional[IPAddress] = None
+    ifindex: Optional[int] = None
+    session_index: Optional[int] = None
+    session_state: Optional[BFDSessState] = None
```
lunkwill42, line 202: For a re-implementation it makes sense to give these more reasonable names, but be aware that these names are taken from the actual API protocol.  If we change these names internally, we must also add aliases to these fields so that they are still exposed under their old names by the server protocol.

src/zino/tasks/bfdtask.py
```diff
@@ -0,0 +1,84 @@
...
+
+    async def run(self):
+        if self._device_state.is_juniper:
+            polled_state = await self._poll_juniper()
+            self._update_state_for_all_ports(polled_state)
+
+    def _update_state_for_all_ports(self, polled_state: BFDStates):
+        for port in self._device_state.ports.values():
+            new_state = polled_state.get(port.ifdescr, BFDState(session_state=BFDSessState.NOSESSION))
+            self._update_state(port, new_state)
```
lunkwill42, line 40: Just a question: Does the original code actually keep a state for ports that do not have any BFD information?

tests/tasks/test_bfdtask.py
```diff
@@ -0,0 +1,97 @@
...
+    device_state = state.devices.get(device_name=device.name)
+    device_state.enterprise_id = 2636
+
+    # this effectively makes the device_port fixture a shortcut for accessing
+    # the port that will change bfd_state when the task is run
+    device_state.ports[device_port.ifindex] = device_port
+    task = BFDTask(device, state)
+    yield task
+
+
```
lunkwill42, line 40: Advocating for the step-down rule, I would want the test code to come first, and the fixtures they depend on to come later on.

An old ref for new devs:
the-stepdown-rule

While it is impossible to adhere 100% to the stepdown rule in a non-compiled language like Python (definitions that are needed at import-time must be defined first), I try to stick to this rule as much as possible, so that I can open any source code file and read the highest-level concepts first.

lunkwill42 (commented):
Not knowing the nitty-gritty details of BFD and the available MIBs, most of this seems just peachy to me :)

I guess we'll still have to run it by Håvard by the end of this project :)

src/zino/statemodels.py
```diff
@@ -29,13 +29,31 @@ class InterfaceState(StrEnum):
...
 
 
+class BFDSessState(Enum):
+    """The set of allowable BFD session states"""
+
+    ADMINDOWN = "adminDown"
+    DOWN = "down"
+    INIT = "init"
+    UP = "up"
+    NOSESSION = "noSession"
```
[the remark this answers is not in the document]
hmpf: "noSession" has never been seen in the wild. Should it be added to zinolib?

src/zino/statemodels.py
```diff
@@ -29,13 +29,31 @@ class InterfaceState(StrEnum):
...
+
+
 class Port(BaseModel):
     """Keeps port state"""
 
     ifindex: int
     ifdescr: Optional[str] = None
     ifalias: Optional[str] = None
     state: Optional[InterfaceState] = None
+    bfd_state: Optional[BFDState] = None
```
hmpf, line 56: This will become a monster of a model, all things optional. The "if"-attributes are only seen on when type is portstate.

hmpf (commented):
See inline comments.

src/zino/tasks/bfdtask.py
stveit: Ill have to look into that. I've mostly disregarded the original implementation since it was known to be based on incorrect assumptions

src/zino/statemodels.py
stveit: Probably not worth the hassle then, I'll change the names back

src/zino/statemodels.py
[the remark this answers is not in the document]
stveit: yes, no session is new just to make it clear that we know that there is no bfd session for that port. Session rows can completely disappear from the MIB, so its valid for a state to go from UP to not existing. The alternative is to just have the session state field be None, but I think its clearer this way.

 Im not fully caught up on zinolib but it might be relevant to add it there

src/zino/tasks/bfdtask.py
```diff
@@ -0,0 +1,84 @@
...
+            session_index = int(index[0])
+            bfd_states[interface_name] = BFDState(
+                session_state=BFDSessState(session_state),
+                session_index=session_index,
+            )
+        return bfd_states
+
+    @property
+    def _device_state(self) -> DeviceState:
+        return self.state.devices.get(self.device.name)
```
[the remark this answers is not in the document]
stveit: #94

stveit:
Had some issue reading ip address from the mib. Turns out you get a Hex-String as response, so you kinda have to parse it yourself. I changed the OctetString parsing in the SNMP module to translate to string using the prettyPrint() method, since the value str() created was difficult to work with. This does not seem to have broken anything, so we might as well use the built-in method meant for converting to strings instead of str()

src/zino/tasks/bfdtask.py
```diff
@@ -0,0 +1,114 @@
...
+            session_discr = row["bfdSessDiscriminator"]
+            session_addr = row["bfdSessAddr"]  # This is a string containing bytes (ex. '\x7f\x00\00x\01')
+            session_addr_type = row["bfdSessAddrType"]
+
+            try:
+                encoded_bytes = str.encode(session_addr, "utf-8")
+                ipaddr = self._convert_address(encoded_bytes, session_addr_type)
+            except ValueError as e:
+                _log.error(f"Error converting bfdSessAddr object to an IP address: {e}")
+                ipaddr = None
```
lunkwill42, line 89: This error log lacks one important piece of information: Which device had the "bad" response?

Testing this branch, I get reams of this error - is that your experience as well?

```
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b"REDACTED" (len 7 != 4) is not permitted as an IPv4 address
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b"REDACTED" (len 6 != 4) is not permitted as an IPv4 address
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b"REDACTED" (len 7 != 4) is not permitted as an IPv4 address
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b"REDACTED" (len 6 != 4) is not permitted as an IPv4 address
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b'REDACTED' (len 5 != 4) is not permitted as an IPv4 address
ERROR - zino.tasks.bfdtask (MainThread) - Error converting bfdSessAddr object to an IP address: b'REDACTED' (len 5 != 4) is not permitted as an IPv4 address
```

I've redacted the actual data, stop by my desk if you need it :)

lunkwill42 (changes requested):
I'm happy with this - except for one tiny nitpick :)

(also, we need to figure out how to tell SonarCloud to ignore mibdumps!)

src/zino/tasks/bfdtask.py
stveit: I can swear this used to work.. Sometimes pysnmp wants to give values that makes sense, sometimes it does not.. 
These errors are caused by for example the hexstring  `80 27 FE 69` gets represented as   `\xc2\x80'\xc3\xbei`, which makes little sense to me.

Changing the parsing of octetstrings to using prettyPrint() instead of just using str() seems to fix it without breaking other stuff. I used prettyPrint() previously, but managed to get it to work without making that change (or so I thought), but it seems to be necessary after all.

lunkwill42 (approved):
No errors now, at least, and state appears to contain valid IP addresses :)

Ignoring SC code smells, since they come from mibdumps

Issue #63, opened 2023-08-15
Implement a Zino 1-compatible API server


lunkwill42:
We must reimplement the text/line-based ("vaguely SMTP-esque") protocol of Zino 1.

The protocol itself is not apparently documented anywhere but in the Zino 1 Tcl code, so this is slightly a form of reverse-engineering it.

Tasks identified so far:

- [x] #89
- [x] #91
- [x] Implement the various API commands from the original Zino protocol
  - [x] #106
  - [x] #93
  - [x] #113
  - [x] #112
  - [x] #99
  - [x] #102
  - [x] #103
  - [x] Implement the `HELP` API command
  - [x] <del>Implement the `NSOCKET` API command</del> (potentially deprecated in Zino 1)
  - [x] #147
  - [x] #218
  - [x] #219
  - [x] Implement the `QUIT` API command
  - [x] #111
  - [x] #101
  - [x] #96
- [x] <del>#90</del>
- [x] <del>#129</del>
- [x] <del>#130</del>
- [x] Verify the API implementation using one of the existing legacy clients (Ritz, cuRitz)

PR #92, opened 2023-10-16
Add challenge/response auth mechanism to the API server protocol

### LINKED ISSUE #91: Write support for the Zino 1 legacy user database and authentication mechanisms

lunkwill42:
Closes #91 .

This does not add all the functionality of the original, as it has the option to configure aspects of authentication through a config file.  We have not yet added a config file feature, so we have added this update to the task list in #63

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -10,6 +10,8 @@ import re
 import textwrap
 from typing import Callable, List, Optional
 
+from zino.api import auth
+
 _logger = logging.getLogger(__name__)
 
 
@@ -28,6 +30,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         self._current_task: asyncio.Task = None
         self._multiline_future: asyncio.Future = None
         self._multiline_buffer: List[str] = []
+        self._authentication_challenge: Optional[str] = None
 
     @property
     def peer_name(self):
@@ -40,7 +43,8 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
     def connection_made(self, transport: asyncio.Transport):
         self.transport = transport
         _logger.debug("New server connection from %s", self.peer_name)
-        self._respond_ok("PLACEHOLDER-CHALLENGE Hello, there")
+        self._authentication_challenge = auth.get_challenge()
+        self._respond_ok(f"{self._authentication_challenge} Hello, there")
 
     def data_received(self, data):
         try:
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -8,7 +8,8 @@ import inspect
 import logging
 import re
 import textwrap
-from typing import Callable, List, Optional
+from pathlib import Path
+from typing import Callable, List, Optional, Union
 
 from zino.api import auth
 
@@ -24,13 +25,15 @@ def requires_authentication(func: Callable) -> Callable:
 class Zino1BaseServerProtocol(asyncio.Protocol):
     """Base implementation of the Zino 1 protocol, with a basic command dispatcher for subclasses to utilize."""
 
-    def __init__(self):
+    def __init__(self, secrets_file: Optional[Union[Path, str]] = "secrets"):
         self.transport: Optional[asyncio.Transport] = None
         self._authenticated: bool = False
         self._current_task: asyncio.Task = None
         self._multiline_future: asyncio.Future = None
         self._multiline_buffer: List[str] = []
         self._authentication_challenge: Optional[str] = None
+
+        self._secrets_file = secrets_file
 
     @property
     def peer_name(self):
@@ -150,7 +153,9 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
         if self.is_authenticated:
             return self._respond_error("already authenticated")
         try:
-            auth.authenticate(user=user, response=response, challenge=self._authentication_challenge)
+            auth.authenticate(
+                user=user, response=response, challenge=self._authentication_challenge, secrets_file=self._secrets_file
+            )
         except auth.AuthenticationFailure as error:
             return self._respond_error(error)
         else:
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -149,11 +149,13 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
         """Implements the USER command"""
         if self.is_authenticated:
             return self._respond_error("already authenticated")
-        if user == "foo" and response == "bar":
-            self._authenticated = True
-            return self._respond_ok("welcome")
+        try:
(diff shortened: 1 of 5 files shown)


src/zino/api/legacy.py
```diff
@@ -22,12 +25,15 @@ def requires_authentication(func: Callable) -> Callable:
 class Zino1BaseServerProtocol(asyncio.Protocol):
     """Base implementation of the Zino 1 protocol, with a basic command dispatcher for subclasses to utilize."""
 
-    def __init__(self):
+    def __init__(self, secrets_file: Optional[Union[Path, str]] = "secrets"):
```
hmpf, line 28: When there is a config-file the location of the "user database" needs to be configurable.

tests/api/auth_test.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+    def test_when_challenge_response_is_bad_should_raise_error(self, secrets_file):
+        with pytest.raises(AuthenticationFailure):
+            assert authenticate(
+                "user1",
+                response="b3e9ba156949152c5a6f3334df86333b1daabfe9",
+                challenge="a97fbebb3eef2dfcf167645229c0c5fa9e92e3da",
+                secrets_file=secrets_file,
+            )
```
hmpf, line 22: These two look identical to me, why should one fail and the other succeed?

hmpf (commented):
Has a test been copy-pasted with insufficient changes?

tests/api/auth_test.py
lunkwill42: Ah, well, `b3e9ba156949152c5c6f3334cf86333b1daabfe9` != `b3e9ba156949152c5a6f3334df86333b1daabfe9`.  Do we need them to be obviously different to the human eye? :)

lunkwill42: alright, new version pushed

tests/api/auth_test.py
```diff
@@ -0,0 +1,59 @@
...
+                response="doesntmatter",
+                secrets_file=secrets_file,
+            )
+
+
+class TestGetChallenge:
+    def test_should_never_return_same_challenge(self):
+        count = 10
+        challenges = set(get_challenge() for _ in range(count))
+        assert len(challenges) == 10, "same challenge was produced more than once"
```
stveit, line 49: This is a weird test imo.

While its unlikely to generate the same value, the `secrets` library does not actually promise to not generate the same value twice. Promising unique values would technically make tokens less secure (or more predictable), as this would mean one possible number is removed every time its run.

The `get_challenge` wrapping function also does nothing to avoid duplicates. So this test technically tests for something that is not actually expected behavior.

Also, this test is technically "flaky", since there is a rare possibility that you actually get duplicates

tests/api/auth_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_should_return_at_least_40_characters(self):
+        assert len(get_challenge()) >= 40
+
+
+def test_read_users_when_given_a_valid_file_then_a_proper_dict_should_be_returned(secrets_file):
+    users = read_users(secrets_file)
+    for user in ("user1", "user2", "user3"):
+        assert user in users
+        assert len(users[user]) == 40
```
stveit, line 59: In one test you check that a challenge is at least 40, and in the next you check that its exactly 40. This is a bit inconsistent, even though it works in this case since you know what the `secret_file` contains. Would be nice to be consistent between these tho for "tests as documentation" purposes, since this kinda implies a requirement for being a "proper" dict is that it contains secrets of exactly length 40.

Feel free to disagree. If this is supposed to test that the secrets in the `secret_file` are returned properly  from `read_users`, then checking the exact length makes some sense. But in that case it might make more sense to check the exact values instead.

The problem here might more be the naming of the test and not the actually testing tho

tests/api/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+@pytest.fixture
+def secrets_file(tmp_path):
+    name = tmp_path.joinpath("secrets")
+    with open(name, "w") as conf:
+        conf.write(
+            """user1 3c55d3cdc19876dfc8c0bb49da8c927a0ddff26d
+            user2 eb6fb35f5fbba6a6e43d3c893ad7a77dc793ceba
+
+            user3 c8a0b250edb2eabd9616b7b05a46e0ad28226fc2
+            """
```
stveit, line 13: is the newline significant?

stveit (changes requested):

tests/api/conftest.py
lunkwill42: No. It's there to ensure that random newlines in the input does not hold any significance to the parser.  This could, of course, be a separate test to make it more explicit.

tests/api/auth_test.py
lunkwill42: > In one test you check that a challenge is at least 40, and in the next you check that its exactly 40. 

No. In one test, I verify that a *challenge issued by Zino* is at least 40 characters. In the next test, I verify that the *secrets* (no challenges are involved here) parsed from the test data are exactly as long as they were in the test data.

More explicitly, they could test that the secrets are identical to the inputs, as this test would not detect that the code changed the secrets into all `A` characters, for example :grin:

stveit: Ah, I see. I've misunderstood the connection between challenge and secrets

tests/api/conftest.py
stveit: yeah i think that intention should be made clearer

tests/api/auth_test.py
lunkwill42: This test is mostly meant to safeguard against a n00b developer b0rking the `get_challenge()` implementation 10 years down the line into something that returns the same static value every time it is called.  It is not meant to be a spec that says that challenges should *NEVER* be the same.

I will personally buy you a very expensive craft beer if you ever observe the secrets library to generate the same value twice in a row during this test :)

stveit: > It is not meant to be a spec that says that challenges should _NEVER_ be the same.

then you should rename the test to not be `test_should_never_return_same_challenge`, Theres not really any other way to interpret "never return same challenge"

stveit (commented):
Also the code smells should be resolved

lunkwill42:
> Also the code smells should be resolved

:-1: The code smells are triggered in files not touched by this PR.  SonarCloud is apparently on drugs...

tests/api/auth_test.py
lunkwill42: done

stveit:
> > Also the code smells should be resolved
> 
> 👎 The code smells are triggered in files not touched by this PR. SonarCloud is apparently on drugs...

lmao

stveit (approved):

PR #95, opened 2023-10-17
Add initial implementation of CASEIDS API command

### LINKED ISSUE #93: Implement the `CASEIDS` API command
The `CASEIDS` API command simply returns a multiline response, in where each line represents  an individual event ID integer for an open event that Zino holds in its state.  Example session:

```
CASEIDS
304 list of active cases follows, terminated with '.'
77049
138951
140038
140039
140041
140042
140215
140216
.
```

lunkwill42:
Closes #93

### DIFF
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -36,9 +36,9 @@ def init_event_loop(args: argparse.Namespace):
     scheduler.add_job(func=state.state.dump_state_to_log, trigger="interval", seconds=30)
 
     loop = asyncio.get_event_loop()
-    echo = loop.create_server(lambda: ZinoTestProtocol(state=state.state), "127.0.0.1", 8001)
-    server = loop.run_until_complete(echo)
-    _log.info("Serving on %r", server.sockets[0].getsockname())
+    server = loop.create_server(lambda: ZinoTestProtocol(state=state.state), "127.0.0.1", 8001)
+    server_setup_result = loop.run_until_complete(server)
+    _log.info("Serving on %r", server_setup_result.sockets[0].getsockname())
     try:
         loop.run_forever()
     except (KeyboardInterrupt, SystemExit):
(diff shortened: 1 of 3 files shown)


src/zino/zino.py
```diff
@@ -36,7 +36,7 @@ def init_event_loop(args: argparse.Namespace):
     scheduler.add_job(func=state.state.dump_state_to_log, trigger="interval", seconds=30)
 
     loop = asyncio.get_event_loop()
-    echo = loop.create_server(lambda: ZinoTestProtocol(), "127.0.0.1", 8001)
+    echo = loop.create_server(lambda: ZinoTestProtocol(state=state.state), "127.0.0.1", 8001)
```
hmpf, line 39: There's `adm_state`, `op_state` and now `zino.state`. Maybe change the latter name, is it for  server state?

Also, what is `echo` used for, is it unfinished?

hmpf (commented):
Something is weird in init_event_loop.

src/zino/zino.py
lunkwill42: `adm_state` and `op_state` are attributes of an event/case.

`zino.state` is a module that holds the global/singleton root state object for the server, correct.  The global state object in that module is also named `state`.  As @stveit pointed out, this does turn into a "state soup" (e.g. `state = state.state`) and can get hard to read.

I've no problem with renames, but we need to agree on a naming system then.  I could live with the module `zino.server_state`, but is the object `zino.server_state.state` any better?

> Also, what is `echo` used for, is it unfinished?

Haha, indeed.  The initial code set up a simple echo server just for exploring how to code a server protocol using asyncio.  I guess the name was left in there by mistake once the code changed.

lunkwill42: Also, I'd say, this is hardly the PR to rename those things, but I will fix the echo-confusion :)

hmpf (approved):

PR #98, opened 2023-10-18
Simplify variable names in API

johannaengland:
As pointed out in https://github.com/Uninett/zino/pull/87/files#r1362053446 the name of the variable `required_args` was different to what it actually stands for.


lunkwill42 (approved):
:+1:

PR #100, opened 2023-10-18
Initial implementation of the GETATTRS API command

### LINKED ISSUE #99: Implement the `GETATTRS` API command
The `GETATTRS` API command takes a case id as input, and returns a list of so-called "simple" case attributes; that is, every event attribute that isn't `log` or `history` (which are usually multiline values).

Example session:

simply returns a multiline response, in where each line represents an individual event ID integer for an open event that Zino holds in its state. Example session:

```
GETATTRS 160279
303 simple attributes follow, terminated with '.'
state: open
router: example-gw3
descr: UX000123 GX42424242, example-oslo-l3
type: portstate
opened: 1697551378
id: 160279
port: xe-0/1/2.0
flaps: 1
updated: 1697551378
priority: 200
polladdr: 192.168.0.142
portstate: up
ifindex: 553
.
```

(clarification: What is internally called an `Event` in Zino is called a `Case` in the outward facing protocol)

lunkwill42:
This adds a new serializer method to the `Event` statemodel, in order to be able to get the event model into a format that matches the original Zino 1 server protocol.

Pydantic seems to provide hooks to write your own serializers for specific fields, but it is unclear to me how this can be utilized when we in fact have two different serialization formats we wish to support (one pure JSON format for persistent state storage, and one legacy server protocol format).  Until someone can point out a better way, this PR includes a new serialization method to `Event` that builds on Pydantic's existing `model_dump()` call.

Closes #99. 

Depends on #95 being merged first.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -195,6 +195,21 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
             self._respond_raw(str(event_id))
         self._respond_raw(".")
 
+    @requires_authentication
+    async def do_getattrs(self, case_id: Union[str, int]):
+        try:
+            case_id = int(case_id)
+            event = self._state.events[case_id]
+        except (ValueError, KeyError):
+            return self._respond_error(f'event "{case_id}" does not exist')
+
+        self._respond(303, "simple attributes follow, terminated with '.'")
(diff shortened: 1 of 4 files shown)


hmpf (approved):

lunkwill42:
#95 was merged to master, rebased this on the latest master.

PR #104, opened 2023-10-18
Initial implementation of GETHIST and GETLOG API commands

### LINKED ISSUE #102: Implement the `GETHIST` API command
The `GETHIST` API command takes a case-id as its argument and responds with a list of all the history entry for the matching event.

Example interactive session from Zino 1:

```
GETHIST 138951
301 history follows, terminated with '.'
1677604216 state change embryonic -> open (monitor)
1677655705 username
 RT 91919191

1677655706 state change open -> waiting (username)
.
```

Each history entry is potentially multi-line.  Each entry begins with a unix timestamp representing the time when the entry was created, followed by a space and then the text in the history entry.  Multiline text is represented more or less like in the SMTP protocol: Each subsequent line belonging to the same entry is prefixed by a space character.

### LINKED ISSUE #103: Implement the `GETLOG` API command
The `GETLOG` API command takes a case-id as its argument and responds with a list of all the log entries for the matching event.

Example interactive session from Zino 1:

```
GETLOG 138951
300 log follows, terminated with '.'
1677604216 ifi2-gw5 peer 2001:7f8:12:1::5:272 AS 50272 is down (running)
.
```

Each log entry is potentially multi-line.  Each entry begins with a unix timestamp representing the time when the entry was created, followed by a space and then the text in the log entry.  Multiline text is represented more or less like in the SMTP protocol: Each subsequent line belonging to the same entry is prefixed by a space character.

`GETHIST` (#102) and `GETLOG` are more or less the same implementation, so they could implemented together.

lunkwill42:
This implements a simple serialization method for `LogEntry` and utilizes this in the implementation of a `GETLOG` and a `GETHIST` protocol implementation.

Closes #102.
Closes #103.
Depends on #100 to merged first.

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -135,6 +135,13 @@ class LogEntry(BaseModel):
     timestamp: datetime.datetime = Field(default_factory=now)
     message: str
 
+    def model_dump_legacy(self) -> List[str]:
+        """Returns the contents of this log entry in the text format expected on the Zino legacy server protocol"""
+        unix_timestamp = int(self.timestamp.timestamp())
+        lines = [f" {line}" for line in self.message.splitlines()]
+        lines[0] = f"{unix_timestamp}{lines[0]}"
+        return lines
+
 
 class EventState(Enum):
     """The set of allowable event states"""
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -136,7 +136,20 @@ class LogEntry(BaseModel):
     message: str
 
     def model_dump_legacy(self) -> List[str]:
-        """Returns the contents of this log entry in the text format expected on the Zino legacy server protocol"""
+        """Returns the contents of this log entry as a sequence of text lines.
+
+        This sequence of text lines are formatted as expected in a multi-line response in the Zino legacy server
+        protocol:
+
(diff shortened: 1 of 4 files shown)


lunkwill42:
All dependencies have been merged to master. Rebased and taken out of draft mode.

src/zino/statemodels.py
```diff
@@ -135,6 +135,13 @@ class LogEntry(BaseModel):
     timestamp: datetime.datetime = Field(default_factory=now)
     message: str
 
+    def model_dump_legacy(self) -> List[str]:
+        """Returns the contents of this log entry in the text format expected on the Zino legacy server protocol"""
```
hmpf, line 139: Instead of "in the text format expected on the Zino legacy server protocol" could you show an example?

hmpf (changes requested):

src/zino/statemodels.py
hmpf: .. otherwise, show examples in the tests so that it will be less necessary to run pdb to see what is passed around later on.

lunkwill42: The tests are already meant to document the expected output, but sure I can update the docstring; see update.

PR #105, opened 2023-10-18
Factor out authenticated protocol object to a test fixture 

lunkwill42:
Many of the tests share the same setup of a pre-authenticated `Zino1ServerProtocol` instance with a mocked transport object with a readable output buffer.  In order to avoid all the redundant test code, this factors out this type of protocol object as a re-usable test fixture.

Depends on #104 to be merged first.


lunkwill42:
#104 is merged now. Rebased to master and taken out of draft mode.

hmpf (approved):
Good refactor.

PR #107, opened 2023-10-19
Add initial ADDHIST API command implementation

### LINKED ISSUE #106: Implement the `ADDHIST` API command

lunkwill42:
It might have been better to break up this PR even further, but the chain of dependent API-related PRs is already becoming too long.  Essentially, these are the changes:

- It changes the `Zino1BaseServerProtocol` to keep track not just of the boolean state of authentication, but also keep track of the username of the logged in user
- It implements the `GETHIST` command, which needs to know said username.
- It factors out the redundant `case_id` verification code that was repeated in all the command responders that take a `case_id` argument.

Closes #106 
Depends on #105 to be merged first.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -249,6 +249,22 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
 
         self._respond_raw(".")
 
+    @requires_authentication
+    async def do_addhist(self, case_id: Union[str, int]):
+        try:
+            case_id = int(case_id)
+            event = self._state.events[case_id]
+        except (ValueError, KeyError):
+            return self._respond_error(f'event "{case_id}" does not exist')
+
+        self._respond(302, "please provide new history entry, terminate with '.'")
+        data = await self._read_multiline()
+        message = f"{self.user}\n" + "\n".join(line.strip() for line in data)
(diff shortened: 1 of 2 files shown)


lunkwill42:
#105 was merged to master, this is now out of draft mode

src/zino/api/legacy.py
```diff
@@ -249,6 +249,22 @@ async def do_getlog(self, case_id: Union[str, int]):
...
+        try:
+            case_id = int(case_id)
+            event = self._state.events[case_id]
+        except (ValueError, KeyError):
+            return self._respond_error(f'event "{case_id}" does not exist')
+
+        self._respond(302, "please provide new history entry, terminate with '.'")
+        data = await self._read_multiline()
+        message = f"{self.user}\n" + "\n".join(line.strip() for line in data)
+        event.add_history(message)
```
hmpf: An example of the correct input (data) and what the message looks like (in a docstring) would be good here, more monday morning friendly.

hmpf (changes requested):
Please document `do_addhist` a bit

src/zino/api/legacy.py
lunkwill42: done

hmpf (approved):

PR #114, opened 2023-10-20
Use list instead of sequence as return type hint

### LINKED ISSUE #74: Make snmp return types consistent
In the snmp.py module there is now  a dataclass `MibObject` and  a named tuple `Identifier`. `getbulk2` and `sparsewalk` returns tuples of (`Identifier`, value), and the others return `MibObject`, that contains an OID and a value. I dont think these should be so separated.

I suggest (if this is possible) to make the return types more consistent. The general return type could be changed to some data structure (or maybe just a normal class) that contains `Identifier` and the value. Something like this:

```
class Identifier(NamedTuple):
    """Identifies a MIB object by MIB, object name and row index"""

    mib: str
    object: str
    index: OID

MibValueTypes = Union[str, int, OID]

class MibObject(NamedTuple):
    """Defines a MIB object with its identifier and value"""

    identifier: Identifier
    value: MibValueTypes

```

Then `MibObject` (or whatever this class should be named) can then be used as return type for all the functions.

Should be discussed if dataclass, NamedTuple or a normal class the most appropriate for this.

stveit:
Fixes #74

According to python 3.9 doc (https://docs.python.org/3.9/library/typing.html#typing.List), abstract type hints like `Sequence` should be used for arguments, while `typing.List` is good for return types. `typing.List` is deprecated in favor of just using `list`, but the point of using something abstract vs using something specific still stands.

So instead of using a mix of `list` and `Sequence`, this PR makes it so we only use `list` for return type hints.

### DIFF
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -187,7 +187,7 @@ class SNMP:
         # var_binds should be a sequence of sequences with one inner sequence that contains the result.
         return var_binds[0][0]
 
-    async def getnext2(self, *variables: Sequence[str]) -> Sequence[SNMPVarBind]:
+    async def getnext2(self, *variables: Sequence[str]) -> list[SNMPVarBind]:
         """Dispatches a GET-NEXT query for the given variables and returns a result where the fetched variables are
         identified symbolically.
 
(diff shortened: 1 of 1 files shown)


lunkwill42 (approved):
Another important effect here is to signal how you expect to be able to treat the value/argument:  Does the function expect to use methods that are only available on lists (like indexing), or are they happy with receiving any kind of sequence-like object (like a generator or something else).

stveit:
> Another important effect here is to signal how you expect to be able to treat the value/argument: Does the function expect to use methods that are only available on lists (like indexing), or are they happy with receiving any kind of sequence-like object (like a generator or something else).

Pretty sure indexing is available for all Sequence objects, but yes this is important.

You dont necessary know exactly how someone is going to use the function in the future, so being specific in what you return makes sense so the "user" of the function knows what is possible. You do however know exactly how your input arguments are used, so here you can be as general as possible.

PR #117, opened 2023-10-20
Add crude test coverage of zino.zino.main()

lunkwill42:
We lack any kind of test coverage of `main()`, which CodeCov complains about the few times we change a line in this function.

This just adds some crude test coverage of main.  This won't detect logic errors, just that the code runs for two seconds without otherwise failing.


lunkwill42:
This doesn't quite work as planned - seems the event loop cannot be restarted/re-used for other tests, so this needs to be re-worked.

lunkwill42:
> This doesn't quite work as planned - seems the event loop cannot be restarted/re-used for other tests, so this needs to be re-worked.

I reworked this to run the Zino "binary" as a subprocess, to avoid the issues with re-using closed event loops.

While my IDE doesn't properly measure coverage in subprocesses without a lot of mucking about with config, our tox/pytest setup can measure coverage in subprocesses just fine, so the coverage published on GitHub will still be correct.

hmpf (approved):

PR #120, opened 2023-10-20
Mock timeouts in SNMP tests

### LINKED ISSUE #118: Speed up tests that wait for timeout
Some tests related to timeouts are "instant", in that they are mocked to raise a `TimeoutError` instead of waiting, but some are not. A few tests in `tests/snmp.test.py` wait for a real timeout, making them slow. Instead, the various snmp commands from pysnmp should be mocked to raise a `TimeoutError`when called.

stveit:
Fixes #118 

Mocks the pysnmp operations `getCmd`, `getBulk` and  `getNext` to give return values that are realistic for when a timeout happens, without actually having to wait for a timeout

### DIFF
--- a/tests/snmp_test.py
+++ b/tests/snmp_test.py
@@ -29,7 +29,7 @@ def unreachable_snmp_client():
     future.set_result(mock_results)
     timeout_mock = Mock(return_value=future)
     with patch.multiple("zino.snmp", getCmd=timeout_mock, nextCmd=timeout_mock, bulkCmd=timeout_mock):
-        device = PollDevice(name="nonexist", address="127.0.0.1", community="invalid", port=666, timeout=1, retries=0)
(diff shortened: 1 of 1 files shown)


tests/snmp_test.py
```diff
@@ -18,10 +22,15 @@ def ipv6_snmp_client(snmpsim, snmp_test_port) -> SNMP:
     return SNMP(device)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture()
```
hmpf, line 25: Why has the scope been removed?

hmpf (commented):
See question.

tests/snmp_test.py
stveit: Per session ended up mocking the pysnmp commands for all tests, not just the ones that using the fixture

lunkwill42 (approved):

PR #121, opened 2023-10-20
Remove fauxpolltask

### LINKED ISSUE #119: Remove FauxPollTask
This is unnecessary now that we have real tasks, and the one test it has is a needless slowdown of the test suite since it just sleeps for a short while.

stveit:
Fixes #119 

Removes fauxpolltask. The only effect fauxpolltask has, it that the test slows down the test suite.


hmpf (approved):
Check this 'un with @lunkwill42 as well in case he still has plans for this thing.

lunkwill42:
> Check this 'un with @lunkwill42 as well in case he still has plans for this thing.

No, not at all.  This was just a placeholder back when we had no real task implementations yet.

lunkwill42 (approved):

PR #122, opened 2023-10-26
Add Cisco support to BFD  Task

### LINKED ISSUE #126: Make BFD task implementation for Cisco

stveit:
Fixes #126


stveit:
force pushed to resolve conflict

hmpf (commented):
Python looks fine, though if we ever have to support a third system for BFD we should probably refactor into system-specific classes.

Standards, what standards *sigh*

lunkwill42:
> Yeah I tried to make something like that, but it ended up feeling very over-engineered. But if we have a 3rd vendor like HP we need to support then it will probably feel better

The rule of thumb I try to follow in these matters:
> Once is by chance. Twice is a coincidence. Three times is a pattern. 

Just my $0.02 :)

lunkwill42 (approved):
It gets kinda clunky once there are two ways to do it, but this passes the smell test for me.  As before, "three times is a pattern".

If we ever get to support this for more than two vendors, things should definitely be refactored.  This also goes for the duplicate tests, which could have benefitted from pytest parametrization to avoid the blatant duplication :)

PR #123, opened 2023-10-27
Add and use device state property to Task

### LINKED ISSUE #94: Add device state property to Task
Discussed in #86, a property like this:

```
    @property
    def _device_state(self) -> DeviceState:
        return self.state.devices.get(self.device.name)

```

should be added to the generic Task class

stveit:
Fixes #94


hmpf (approved):
Such a little change, such a great boon in readability!

PR #128, opened 2023-11-28
Revert "correct" handling of SNMP errors

stveit:
Reverts changes made in #116. @johannaengland found that sometimes the `var_bind` return from pysnmp commands contain `ObjectTypes` where `objecttype[0]` is an error message and not an OID as expected .. Since  pysnmp wants to be unpredictable its better to just do it the previous way that seemed to work better.

### DIFF
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -133,7 +133,7 @@ class SNMP:
         error_indication: Union[str, errind.ErrorIndication],
         error_status: str,
         error_index: int,
-        var_binds: Sequence[ObjectType],
+        query: Sequence[ObjectType],
     ):
         """Raises a relevant exception if an error has occurred"""
         # Local errors (timeout, config errors etc)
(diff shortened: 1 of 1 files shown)


stveit:
> I don't think this is pysnmp being weird. This is probably just a result of how SNMP v2c works. I think I recently made some similar comments on how errors were handled / not handled in NAV during the recent SNMP v3 additions there.
> 
> SNMP v2c introduces a concept of "exceptions" being raised by an SNMP agent, and these "exceptions" are flagged as specific return value types in the varbinds of a response PDU (SNMP v1 only had the error-flag and error-index way of reporting errors back to the querying entity).
> 
> In practice, this means that if you send an SNMP request with multiple varbinds in it (i.e. you're querying multiple objects in the same request), and the responding agent encountered an error with one of those varbinds, it can still return a response that contains a value for the ok varbinds and an error for the problematic varbinds.
> 
> I don't think I have the specifics on what actually failed in your case...

I see. In other words, back to the drawing board. I reoponed the original issue, so a potential future PR that takes this stuff into consideration can reference to that. I still think this revert PR should be merged, since it will probably take some time to figure out a proper fix

lunkwill42:
Sure, we can revert until we have a better fix. The original attempt will still be here, and could be worked on further.

PR #131, opened 2023-11-30
Pretty print IP addresses in SNMP

johannaengland:


lunkwill42 (approved):
I don't like the name "Pretty Print", as we're not printing anything here.  We can't change the PySNMP API, but I changed the title of the PR to make it more clear what this is all about :)

PR #133, opened 2024-01-03
Bgp state monitor

### LINKED ISSUE #64: Make a BGP state monitor task

stveit:
Fixes #64 

Contains code from @johannaengland 

I added some tests and scrubbing of snmprec files.

The tests are very specific in their naming. I did this because they are very much based on the code itself, so if there is something wrong with the code some of the tests might be bad also. For example
`test_peer_admin_status_changing_to_stop_while_peer_state_is_not_established_should_create_event_general` is extremely specific and lengthy. Because I was unsure about the overall intent of the code I described it like this so its clear what it is based on. From reading the code, I could see that admin-down events were created when admin status was `stop `specifically while peer_state was not"`established`.  A question I had was what should happen if admin status is"`stop` but peer-state is still `established`? Should this not create an admin-down event? Or is this something that is impossible to begin with?

Basically I need some more insight in the intention behind the code to finalize these things.

I also found a bug that I patched in 532357dfec2616a1e0eac009d0b6dfc0fc2d5474.
From what I know this code is pretty much 1 to 1 with the original tcl code, so I'm wondering if this is also in that code or just a mistake done while porting (or maybe its not a bug at all?). @johannaengland can probably shed some light on this.


stveit:
The tests ran mostly fine locally, but seems there is an issue with the MIB dumps when running here. I did have an error locally that always happened to the last test that was run, so its not something tied to the individual tests

johannaengland (commented):
I will have a look at why the tests are failing

hmpf (commented):
What logging statements were you wondering about?

johannaengland:
> What logging statements were you wondering about?

Just some examples:
https://github.com/Uninett/zino/pull/133/files#diff-1e7cb9d466f6ea6dbc926470743ae453bd651952c0dff24b24c73de4538e2c6fR307-R308 and
https://github.com/Uninett/zino/pull/133/files#diff-1e7cb9d466f6ea6dbc926470743ae453bd651952c0dff24b24c73de4538e2c6fR371-R376

Or 
https://github.com/Uninett/zino/pull/133/files#diff-1e7cb9d466f6ea6dbc926470743ae453bd651952c0dff24b24c73de4538e2c6fR325-R327 and
https://github.com/Uninett/zino/pull/133/files#diff-1e7cb9d466f6ea6dbc926470743ae453bd651952c0dff24b24c73de4538e2c6fR413-R418

stveit (commented):
This seems fine to me. Might steal the pytest parametrizartion for BFD.
I can do some more manual testing for this tomorrow.

johannaengland:
Another thing that I am also uncertain about are the log levels - Zino 1.0 doesn't really have different levels, so some input when which level would be appropriate would be great

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,420 @@
...
+        try:
+            entry = await snmp.getnext("BGP4-MIB", "bgp")
+        except NoSuchNameError:
+            pass
+        else:
+            if entry and _is_in_scope(entry=entry, oid=("BGP4-MIB", "bgp")):
+                return "general"
+
+        return None
+
```
lunkwill42, line 93: Following the rule of "three times is a pattern", this should be boiled down to something simpler, and the suggestion for making a generic test-for-support method would help in that regard

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,420 @@
...
+                return None
+
+    async def _get_uptime(self) -> Union[int, None]:
+        snmp = SNMP(self.device)
+        try:
+            return (await snmp.get("SNMPv2-MIB", "sysUpTime", 0)).value / 100
+        except NoSuchNameError:
+            _logger.info(f"router {self.device.name} misses sysUpTime")
+            return None
+
```
lunkwill42, line 123: The `LinkStateTask` does more or less the same thing.. here's an opportunity to make some common function out of it...

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,425 @@
...
+            ("peer_fsm_established_time", "cbgpPeer2FsmEstablishedTime"),
+        ]
+        general_translation = [
+            ("peer_state", "bgpPeerState"),
+            ("peer_admin_status", "bgpPeerAdminStatus"),
+            ("peer_remote_address", "bgpPeerRemoteAddr"),
+            ("peer_remote_as", "bgpPeerRemoteAs"),
+            ("peer_fsm_established_time", "bgpPeerFsmEstablishedTime"),
+        ]
+
```
lunkwill42, line 229: These maps look like constants to me (with some redundancies that could perhaps be factored out) - why redeclare them on every call to this method?

lunkwill42, line 229: These maps look like constant values, so why re-define them on every call to this function?

Also, in combination with the following if-else statement, they could be made somewhat more compact...

lunkwill42 (changes requested):
So, I've been trying to do the review from my IDE, but some of my comments seem to be disappearing - and currently it says I have 9 pending comments, but I can't see them all, so I suspect there may be some duplicates occurring in the actual review.  Sorry about that.


I'll probably have more to say about this later, but this will have to do for now.  I'm sure this follows the Tcl code pretty well, and does the job, but I do see ample opportunity for factoring out some generic patterns that keep occurring througout.

My tip about replacing `Union[something, None]` with `Optional[something]` could be repeated multiple times, but you can find all the instances yourself :)

SonarCloud does flag a refactoring opportunity for `_update_single_bgp_entry`, which I think definitely has merit.  This method is just too darn large...

Great work!

src/zino/tasks/bgpstatemonitortask.py
johannaengland: Where would I best move that function to?

src/zino/tasks/bgpstatemonitortask.py
johannaengland: Good catch!

johannaengland:
I have now included all of the requested changes except for the `sysUptime` one, waiting for an answer there. 

I also would appreachiate some tips on how to refactor `_update_single_bgp_entry`.

src/zino/snmp.py
```diff
@@ -388,6 +388,18 @@ def _var_bind_is_in_scope(var: PySNMPVarBind) -> bool:
                 break  # Nothing left to query
         return dict(results)
 
+    def _is_in_scope(self, entry: MibObject, oid: Tuple[str, str]):
+        object_type = self._oid_to_object_type(*oid)
+        self._resolve_object(object_type=object_type)
+        root = OID(object_type[0])
+        return root.is_a_prefix_of(other=entry.oid)
+
+    async def subtree_is_supported(self, *oid: str) -> bool:
```
lunkwill42, line 397: This should have a docstring that explains the purpose of this method.

src/zino/snmp.py
```diff
@@ -388,6 +388,18 @@ def _var_bind_is_in_scope(var: PySNMPVarBind) -> bool:
                 break  # Nothing left to query
         return dict(results)
 
+    def _is_in_scope(self, entry: MibObject, oid: Tuple[str, str]):
```
lunkwill42, line 391: You are calling this method from another class, which means it should probably not marked as private.  I suggest removing the leading underscore, and adding a docstring that explains its purpose/usage.

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,410 @@
...
+                _logger.info(f"router {self.device.name} misses bgpLocalAs")
+                return None
+
+    async def _get_uptime(self) -> Optional[int]:
+        snmp = SNMP(self.device)
+        try:
+            return (await snmp.get("SNMPv2-MIB", "sysUpTime", 0)).value / 100
+        except NoSuchNameError:
+            _logger.info(f"router {self.device.name} misses sysUpTime")
+            return None
```
lunkwill42, line 125: Given a constant defined at the module-level as

```python
LOCAL_AS_OBJECTS = {
    "juniper": ("BGP4-V2-MIB-JUNIPER", "jnxBgpM2PeerLocalAs"),
    "cisco": ("CISCO-BGP4-MIB", "cbgpLocalAs"),
    "general": ("BGP4-MIB", "bgpLocalAs")
}
```

This method could be simplified as:


```suggestion
    async def _get_local_as(self, bgp_style: BgpStyle) -> Optional[int]:
        snmp = SNMP(self.device)
        try:
            mib, object_name = LOCAL_AS_OBJECTS[bgp_style]
        except KeyError:
            return

        response = await snmp.getnext(mib, object_name)
        if snmp._is_in_scope(response, (mib, object_name)):
            return response.value
        else:
            _logger.info(f"router {self.device.name} misses {object_name}")
```

src/zino/tasks/bgpstatemonitortask.py
lunkwill42: Good question. I might say that you don't need to fetch sysUptime, as it has already been fetched by the linkstatetask immediately prior to this, and added to the device state as the boot time.

But, then again, in some cases you might need an accurate and up-to-date value.  It could be argued that `_get_uptime()` would be such a common operation that maybe it should even be in the `Task` base class itself?

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,410 @@
...
+                        )
+
+        # Update all states
+        self.device_state.bgp_peer_oper_states[data.peer_remote_address] = data.peer_state
+        self.device_state.bgp_peer_admin_status[data.peer_remote_address] = data.peer_admin_status
+        self.device_state.bgp_peer_up_times[data.peer_remote_address] = data.peer_fsm_established_time
+
+        # Add new peers if found
+        if data.peer_remote_address not in self.device_state.bgp_peers:
+            self.device_state.bgp_peers.append(data.peer_remote_address)
```
lunkwill42, line 332: You asked for suggestions for a refactor.  Without diving very deep into the logic here, it could at least be easily broken down into multiple methods using a simple refactor operation (usually even available as automated tools from your IDE).

One very quick suggestion is first to extract this into a module-level constant:
```python
BUGGY_REMOTE_ADDRESSES = [
    # Bug in JunOS -- info from IPv6 BGP sessions spill over
    "0.0.0.0",
    # Bug in earlier Cisco IOS, info from elsewhere (IPv6?) spills over
    "32.1.7.0",
]
```

Then just factor out some of the branches like so:

```suggestion
    def _update_single_bgp_entry(self, oid: OID, row: dict[str, Any], local_as: int, uptime: int):
        data = BaseBgpRow(**row)
        if data.peer_remote_address in BUGGY_REMOTE_ADDRESSES:
            return

        # Internal bgp sessions are not observed
        if local_as == data.peer_remote_as:
            return

        if data.peer_state == "established":
            self._update_established_peer(data, uptime, oid)
        else:
            self._update_nonestablished_peer(data, uptime)

        # Update all states
        self.device_state.bgp_peer_oper_states[data.peer_remote_address] = data.peer_state
        self.device_state.bgp_peer_admin_status[data.peer_remote_address] = data.peer_admin_status
        self.device_state.bgp_peer_up_times[data.peer_remote_address] = data.peer_fsm_established_time

        # Add new peers if found
        if data.peer_remote_address not in self.device_state.bgp_peers:
            self.device_state.bgp_peers.append(data.peer_remote_address)

    def _update_established_peer(self, data, uptime, oid):
        index = f"{self.device.name},{oid}"

        saved_bgp_peer_up_time = self.device_state.bgp_peer_up_times.get(data.peer_remote_address, None)
        if saved_bgp_peer_up_time and uptime >= saved_bgp_peer_up_time > data.peer_fsm_established_time:
            self._bgp_external_reset(data)
            _logger.debug(f"Noted external reset for {self.device_state.name}: {index}")
        else:
            event = self.state.events.get(self.device.name, data.peer_remote_address, BGPEvent)
            if event and event.operational_state != "established":
                self._bgp_external_reset(data)
                _logger.debug(f"BGP session up for {self.device_state.name}: {index}")

    def _update_nonestablished_peer(self, data, uptime):
        saved_bgp_peer_admin_status = self.device_state.bgp_peer_admin_status.get(data.peer_remote_address, None)
        if not saved_bgp_peer_admin_status:
            self.device_state.bgp_peer_admin_status[data.peer_remote_address] = "unknown"
            saved_bgp_peer_admin_status = "unknown"
        if data.peer_admin_status in ["stop", "halted"]:
            if saved_bgp_peer_admin_status != data.peer_admin_status:
                self._bgp_admin_down(data)
                _logger.debug(
                    f"Router {self.device_state.name} peer {data.peer_remote_address} AS {data.peer_remote_as} "
                    f"admin-down"
                )
        else:
            self._update_peer_with_admin_status_start_or_running(data, saved_bgp_peer_admin_status, uptime)

    def _update_peer_with_admin_status_start_or_running(self, data, saved_bgp_peer_admin_status, uptime):
        if saved_bgp_peer_admin_status != data.peer_admin_status:
            self._bgp_admin_up(data)
        saved_bgp_peer_oper_state = self.device_state.bgp_peer_oper_states.get(data.peer_remote_address, None)
        if not saved_bgp_peer_oper_state:
            self.device_state.bgp_peer_oper_states[data.peer_remote_address] = "established"
            saved_bgp_peer_oper_state = "established"
        if saved_bgp_peer_oper_state == "established":
            # First verify that we've been up longer than the required time before we flag it as an alert
            if uptime > TIME_BEFORE_OPER_DOWN_ALERT:
                self._bgp_oper_down(data)
                _logger.debug(
                    f"Router {self.device_state.name} peer {data.peer_remote_address} AS {data.peer_remote_as} "
                    f"is {data.peer_state} (down)",
                )
            else:
                _logger.debug(
                    f"Router {self.device_state.name} peer {data.peer_remote_address} AS {data.peer_remote_as} "
                    f"is {data.peer_state} (down), but uptime = {uptime}",
                )
```

The `oid` argument to `_update_single_bgp_entry` appears to be there exclusively for constructing some debug log statements, which looks a bit weird.  Maybe it could be obtained some other way, but that would possibly require a deeper analysis than I've committed to :)

lunkwill42 (changes requested):
Some more feedback per your own comments/questions.

Also, I have an itch when it comes to the repeated use of this in many methods:

```python
snmp = SNMP(self.device)
```

It seems wasteful to establish multiple SNMP sessions across the lifetime of a single task.  This is the point where you should consider adding this to the `run()` method:

```python
self.snmp = SNMP(self.device)
```

Then re-use the same SNMP session across the other methods.

src/zino/tasks/bgpstatemonitortask.py
johannaengland: Very good catch, with the oid, I realized it was a mistake on my part and I used the wrong value for logging

johannaengland:
After figuring out that the oid is not needed for any kind of logging purpose I realized that for devices of BGP style juniper and general we don't need them at all and for cisco only to extract the `peer_remote_address` from it. Should I try to rework it so that after extracting `peer_remote_address` for cisco we don't continue carrying the oid along and instead just saving all entries in a list or should I keep it the way it is currently?

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,393 @@
...
+        uptime = await self._get_uptime()
+        if not local_as or not uptime:
+            return
+
+        if bgp_style == "juniper":
+            bgp_info = await self._get_juniper_bgp_info()
+        elif bgp_style == "cisco":
+            bgp_info = await self._get_cisco_bgp_info()
+        elif bgp_style == "general":
+            bgp_info = await self._get_general_bgp_info()
```
hmpf, line 83: There is enough repetition here that it is tempting to make a class per type (cisco/juniper/generic) and put the classes in a dict, so we get 

```
bgp_info = BGP_TYPES[bgp_style]._get_bgp_info()
```

Probably better to wait with that though, at least until we know everything works correctly.

hmpf (commented):
This would have been better as two PRs: one to set `self._snmp` in task, and one to add the bpg support. I'm not saying you should split it, just for next time.

It seems that every method name in *Task except "run" starts with an underscore and that, to me, is.. infelicitous? noisy? code smell for Python? Don't quite know. Don't change this either, it just feels like extra sand in the eyes on this Monday morning.

src/zino/statemodels.py
```diff
@@ -109,7 +113,7 @@ def set_boot_time_from_uptime(self, uptime: int):
 
         :param uptime: An uptime value in 100ths of a second
         """
-        self.boot_time = now() - datetime.timedelta(seconds=uptime / 100)
+        self.boot_time = now() - datetime.timedelta(seconds=uptime)
```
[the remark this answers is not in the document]
johannaengland: I will also separate this into an extra pull request to make this one less complex

lunkwill42:
> It seems that every method name in *Task except "run" starts with an underscore and that, to me, is.. infelicitous? noisy? code smell for Python? Don't quite know. Don't change this either, it just feels like extra sand in the eyes on this Monday morning.

I don't see this as a problem.  None of these methods are meant for public usage outside of the class, so the idiomatic way to signal this in Python is by prefixing their names with an underscore.

src/zino/statemodels.py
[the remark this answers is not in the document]
lunkwill42: Ah, this is where the reduction of precision came about in #153 :grin: 

This is in fact just fine for the timdelta constructor, i.e. it breaks down a float intput like this, @hmpf  :)

```pycon
>>> from datetime import timedelta
>>> timedelta(seconds=50.5)
datetime.timedelta(seconds=50, microseconds=500000)
>>>

**Review by hmpf — APPROVED** [review]
Wait for @lunkwill42's ack before you merge.

**Comment by hmpf** [comment]
> the idiomatic way to signal this in Python is by prefixing their names with an underscore.

Python isn't very idiomatic though.

**Comment by lunkwill42** [comment]
> Python isn't very idiomatic though.

That's just like, your opinion, man! ;-)

File: src/zino/statemodels.py [hunk 10]
```diff
@@ -72,6 +75,11 @@ class DeviceState(BaseModel):
     ports: Dict[int, Port] = {}
     alarms: Optional[Dict[AlarmType, int]] = None
     boot_time: Optional[datetime.datetime] = None
+    bgp_style: Optional[BgpStyle] = None
+    bgp_peers: Optional[List[IPAddress]] = []
+    bgp_peer_up_times: Optional[Dict[IPAddress, int]] = dict()
+    bgp_peer_admin_status: Optional[Dict[IPAddress, BgpAdminStatus]] = dict()
+    bgp_peer_oper_states: Optional[Dict[IPAddress, BgpOperState]] = dict()
```
lunkwill42, line 82: I find it hard to resist commenting on the data structures this time around, but I think this note should be moved forward to a future refactor:

We don't live in the flat-dict world of Zino 1 / Tcl any more: This is a bunch of dicts with the same index.  This would probably play better if `BGPPeerSession` was a separate model of attributes concerning a peering session, and then `bgp_peers` would be the single `dict` on `Device` that indexes them.

lunkwill42: > We don't live in the flat-dict world of Zino 1 / Tcl any more: This is a bunch of dicts with the same index. This would probably play better if `BGPPeerSession` was a separate model of attributes concerning a peering session, and then `bgp_peers` would be the single `dict` on `Device` that indexes them.

Of course, if we push ahead with an alpha release, and change this before the final release, we have to think about whether we want to support migrating from older state data, or just declare that the format isn't going to be stable until a final release.

src/zino/statemodels.py
```diff
@@ -75,11 +109,8 @@ class DeviceState(BaseModel):
     ports: Dict[int, Port] = {}
     alarms: Optional[Dict[AlarmType, int]] = None
     boot_time: Optional[datetime.datetime] = None
-    bgp_style: Optional[BgpStyle] = None
-    bgp_peers: Optional[List[IPAddress]] = []
-    bgp_peer_up_times: Optional[Dict[IPAddress, int]] = dict()
-    bgp_peer_admin_status: Optional[Dict[IPAddress, BgpAdminStatus]] = dict()
-    bgp_peer_oper_states: Optional[Dict[IPAddress, BgpOperState]] = dict()
+    bgp_style: Optional[BGPStyle] = None
+    bgp_peers: Optional[dict[IPAddress, BGPPeerSession]] = dict()
```
hmpf, line 113: "Optional" means that `bgp_peers` may be `None`, but in the same line you set a default of "dict()" which means it doesn't start out as None. Does a value "None" make sense, considering that `bool(dict()) is bool(None)`, that is: both falsey?

src/zino/statemodels.py
```diff
@@ -13,13 +13,39 @@
...
+class BGPAdminStatus(StrEnum):
+    """The set of allowable BGP admin status"""
+
+    RUNNING = "running"
+    HALTED = "halted"
+    START = "start"
+    STOP = "stop"
+
+
+class BGPOperStates(StrEnum):
```
hmpf, line 37: To pluralize or not to pluralize, that's the question.

lunkwill42: Yes, good point. An instance of this represent a single state, not multiple, so the class should really have a non-plural name

tests/tasks/test_bgpstatemonitortask.py
```diff
@@ -192,6 +192,30 @@ async def test_oper_down_creates_event(self, task):
...
+        # set initial state
+        if task.device_state.bgp_style == "juniper":
+            task.device_state.bgp_peer_admin_status = {peer_address: "running"}
+        else:
+            task.device_state.bgp_peer_admin_status = {peer_address: "start"}
+        task.device_state.bgp_peer_oper_states = {peer_address: "established"}
+        await task.run()
+        # check if state has been updated to reflect state defined in .snmprec
+        assert task.device_state.bgp_peer_oper_states[peer_address] != "established"
+        # check that the correct event has been created
```
lunkwill42: Surely you mean:
> Verify that no event was created

?

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,404 @@
...
+            return
+        response = await self.snmp.getnext(mib, object_name)
+        if self.snmp.is_in_scope(response, (mib, object_name)):
+            return response.value
+        else:
+            _logger.info(f"router {self.device.name} misses {object_name}")
+
+    async def _get_uptime(self) -> Optional[int]:
+        try:
+            return (await self.snmp.get("SNMPv2-MIB", "sysUpTime", 0)).value / 100
```
stveit, line 129: The parent Task class has a built-in _get_uptime method that can be used (dont know if that this PR has that update uncluded)

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -94,6 +94,10 @@ async def run(self):
 
         bgp_info = self._fixup_ip_addresses(bgp_info=bgp_info)
 
+        # BGP peers is by default None, so to record the BGP peers, we have to initialize it
+        if not self.device_state.bgp_peers:
+            self.device_state.bgp_peers = dict()
+
```
lunkwill42, line 100: Actually, I would go the other way around and make `DeviceState.bgp_peers` non-optional with a default value of `{}`, similar to `DeviceState.ports`.  You save three lines of unnecessary init-code that may have to be repeated elsewhere when things change in the future.

lunkwill42 (changes requested):
Thank you for that clean-up, it made the code a lot easier to read and reason about :)

Still have some nitpicks with some of the changes, so here we go....

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,404 @@
...
+        uptime = await self._get_uptime()
+        if not local_as or not uptime:
+            return
+
+        if bgp_style == BGPStyle.JUNIPER:
+            bgp_info = await self._get_juniper_bgp_info()
+        elif bgp_style == BGPStyle.CISCO:
+            bgp_info = await self._get_cisco_bgp_info()
+        elif bgp_style == BGPStyle.GENERAL:
+            bgp_info = await self._get_general_bgp_info()
```
stveit, line 90: feels a bit inconsistent that `_get_local_as` can take bgp_style as a parameter while you use if-else statements for deciding which `get_vendor_bgp_info` to use.

It could be useful to make separate classes for handling Juniper/cisco/general that each implement a `get_bgp_info` and `get_local_as`, basically delegating elsewhere the responsibility to decide which style should be used with a factoru pattern or something. The overhead might not be worth it tho.

stveit (commented):
The new models are nice, and I like the way the BFD logic is split up more into separate funtions, makes it much easier to see whats going on. 

Got some inline comments, but most of it is just me thinking loudly

tests/tasks/test_bgpstatemonitortask.py
johannaengland: I think you looked at an old commit, this was fixed in https://github.com/Uninett/zino/pull/133/commits/c8f2f4ffba2f921ee14c2dc021486677addca8d3

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,404 @@
...
+            return response.value
+        else:
+            _logger.info(f"router {self.device.name} misses {object_name}")
+
+    async def _get_uptime(self) -> Optional[int]:
+        try:
+            return (await self.snmp.get("SNMPv2-MIB", "sysUpTime", 0)).value / 100
+        except NoSuchNameError:
+            _logger.info(f"router {self.device.name} misses sysUpTime")
+            return None
```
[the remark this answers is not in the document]
johannaengland: Oh, forgot to remove it here, good catch!

src/zino/tasks/bgpstatemonitortask.py
johannaengland: fixed

src/zino/tasks/bgpstatemonitortask.py
johannaengland: Yes, that idea has also been raised by @hmpf and has been included in #158 for the future.

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -0,0 +1,404 @@
...
+        if await self.snmp.subtree_is_supported("BGP4-V2-MIB-JUNIPER", "jnxBgpM2"):
+            return BGPStyle.JUNIPER
+
+        if await self.snmp.subtree_is_supported("CISCO-BGP4-MIB", "cbgpPeer2Table"):
+            return BGPStyle.CISCO
+
+        if await self.snmp.subtree_is_supported("BGP4-MIB", "bgp"):
+            return BGPStyle.GENERAL
+
+        return None
```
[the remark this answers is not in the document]
johannaengland: I will also add this to #158

hmpf (approved):

hmpf (approved):
Can `self.device.address` and `self.device.priority` ever change? If it can't they should only ever be set in `get_or_create_event`. Maybe something to add for #158.

johannaengland:
> Can `self.device.address` and `self.device.priority` ever change? If it can't they should only ever be set in `get_or_create_event`. Maybe something to add for #158.

But then that would be a general issue, not just for BGP, so I feel it deserves its own issue instead of being added to #158.

lunkwill42 (approved):
Let's go!

lunkwill42:
> > Can `self.device.address` and `self.device.priority` ever change? If it can't they should only ever be set in `get_or_create_event`. Maybe something to add for #158.
> 
> But then that would be a general issue, not just for BGP, so I feel it deserves its own issue instead of being added to #158.

It is a general issue, as it seems Zino 1 does this for everything.  However (and we should check with Håvard), a router could change its address or priority during the lifetime of any event that relates to it, so I guess it *could* change.  But whether it matters is a different question (for Håvard).

Nevertheless, the BGP task itself does have a lot of internal redundancies when it comes to creating events (I didn't realize it actually did it in so many places) - and that bit alone could certainly be considered for refactoring in #158.

johannaengland:
> Nevertheless, the BGP task itself does have a lot of internal redundancies when it comes to creating events (I didn't realize it actually did it in so many places) - and that bit alone could certainly be considered for refactoring in #158.

Yes, that I agree with and had noticed it as well when adding the change for priority and polladdr. I will add it to #158.

PR #136, opened 2024-01-09
Add an event commit mechanism to the event registry

### LINKED ISSUE #52: Add mechanism to commit event changes
Issue of the inline comment 

`# TODO we need a mechanism to "commit" event changes, to trigger notifications to clients`

Cleaner to have it as an issue and not comment

lunkwill42:
This is a work-in-progress to add a checkout/commit mechanism for events in the event state registry.

The mechanism enables "checking out" copies of event objects, which can subsequently changed and "committed" back to the event state registry.

This also adds an observer mechanism, so that observers can be notified about events being committed.  Observers can be used to trigger imminent state dumps or notifications to API clients that have active notification channels open.

Closes #52

### DIFF
--- a/src/zino/events.py
+++ b/src/zino/events.py
@@ -1,6 +1,6 @@
 import logging
 from collections import namedtuple
-from typing import Dict, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
 
 from pydantic.main import BaseModel
 
@@ -25,6 +25,7 @@ class Events(BaseModel):
     events: Dict[int, Union[PortStateEvent, BGPEvent, BFDEvent, ReachabilityEvent, AlarmEvent, Event]] = {}
     last_event_id: int = 0
     _events_by_index: Dict[EventIndex, Event] = {}
+    _observers: list[Callable[[int], Any]] = []
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -114,6 +115,16 @@ class Events(BaseModel):
         self.events[event.id] = event
         if is_new:
             self._rebuild_indexes()
+
+        self._call_observers_for(event)
+
+    def add_event_observer(self, func: callable):
+        """Adds an observer function that will be called with the ID of any committed event as its argument"""
+        self._observers.append(func)
+
+    def _call_observers_for(self, event: Event):
+        for observer in self._observers:
+            observer(event.id)
 
 
 class EventExistsError(Exception):
--- a/src/zino/events.py
+++ b/src/zino/events.py
@@ -104,8 +104,8 @@ class Events(BaseModel):
     def commit(self, event: Event):
         """Commits an Event object to the state, replacing any existing event by the same id.
 
-        If the event, for some reason, does not replace an existing event, indexes are rebuilt. This entrusts the
-        committer to not change the identifying index attributes of a modified event.
+        If the event, for some reason, does not replace an existing event, indexes are rebuilt. This assumes the
+        committer does not change the identifying index attributes of a modified event.
         """
         if event.state == EventState.EMBRYONIC:
             event.state = EventState.OPEN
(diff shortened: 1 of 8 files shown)


lunkwill42:
CodeCov complains about missing coverage in two changed lines of `zino.zino.main`.  I don't consider adding coverage to the main method part of this PR, it is already being worked on in #117

src/zino/events.py
```diff
@@ -96,6 +97,35 @@ def get(self, device_name: str, port: Optional[PortOrIPAddress], event_class: Ty
...
         return self._events_by_index.get(index)
 
+    def checkout(self, event_id: int) -> Event:
+        """Checks out a copy of an event that can be freely modified without being persisted"""
+        return self[event_id].model_copy(deep=True)
+
+    def commit(self, event: Event):
+        """Commits an Event object to the state, replacing any existing event by the same id.
+
+        If the event, for some reason, does not replace an existing event, indexes are rebuilt. This entrusts the
```
hmpf, line 107: Clunky sentence. Maybe "This assumes the committer does not change the identifying index attributes of a modified event."

hmpf (commented):
Some *twitch-twitch* from bare integers floating around as timeouts. See also inline comment.

lunkwill42:
> Some _twitch-twitch_ from bare integers floating around as timeouts. See also inline comment.

Timeouts? What are you referring to?

src/zino/events.py
lunkwill42: Agreed. Fixed.

lunkwill42:
As mentioned in this morning's standup meeting, I'm not quite sure how this will play into the `get_or_create_event()` method.  Most task plugins will use this to get an event to modify when changes are detected.  However, currently, this returns an `Event` object directly from the events list, which means the task plugin is modifying the "live" object and later committing it.

In all probability, this method should also be changed to use the checkout mechanism for events.  Which, as commented during the meeting, should probably also be documented very specifically in its docstring:  The returned object must be committed for any changes to take effect.

hmpf:
> > Some _twitch-twitch_ from bare integers floating around as timeouts. See also inline comment.
> 
> Timeouts? What are you referring to?

`def reschedule_dump_state_on_commit(event_id: int, max_wait=timedelta(seconds=10)):` and `scheduler.add_job(func=state.state.dump_state_to_log, trigger="interval", seconds=30)`. The "10" is a timeout, yes? Dunno about the other one. The first one screams to be a config variable eventually.

lunkwill42:
> `def reschedule_dump_state_on_commit(event_id: int, max_wait=timedelta(seconds=10)):` and `scheduler.add_job(func=state.state.dump_state_to_log, trigger="interval", seconds=30)`. The "10" is a timeout, yes? Dunno about the other one. The first one screams to be a config variable eventually.

That's not how I think of it, @hmpf, but I guess you could construe it that way, if you want.  It is a limit that serves two purposes: If the next scheduled dump job is more than 10 seconds away, we want to reschedule it to run in exactly 10 seconds.  If it's 10 seconds or less until the next job, skip the rescheduling and let things run their course.

In any case, the value is in fact hardcoded in the legacy Tcl code, so I didn't bother to do much about it here, other than accepting an optional argument for it.  It could also be a global constant, of course.

lunkwill42:
> `scheduler.add_job(func=state.state.dump_state_to_log, trigger="interval", seconds=30)`.

And, as discussed elsewhere, this job should be removed. It was put there for debugging purposes before we had a working dump-do-disk mechanism.  The state can get quite large, too large for logs, really.

johannaengland (approved):

lunkwill42:
> This isn't part of this PR anyway. And, as discussed elsewhere, this job should be removed. It was put there for debugging purposes before we had a working dump-to-disk mechanism. The state can get quite large, too large for logs, really.


Added #140 to clean up this.

src/zino/zino.py
```diff
@@ -11,7 +11,7 @@
 from zino.config.models import DEFAULT_INTERVAL_MINUTES
 from zino.scheduler import get_scheduler, load_and_schedule_polldevs
 
-STATE_DUMP_JOB_ID = "zino.dumpo_state"
```
johannaengland: :laughing:

PR #138, opened 2024-01-10
Use checkout/commit logic for all event creation/retrieval from the index

### LINKED ISSUE #137: Consolidate logic for switching an event's state from `EMBRYONIC` to `OPEN`
Currently, any piece of code that creates a new Zino event will explicitly set the initial event state and add a log entry for the opening of the event, resulting in what would seem to be unnecessary redundancies.

Once #136 is merged, this responsibility should be moved.  The event registry `commit()` method of this #136 will automatically switch an event's state from `EMBRYONIC` to `OPEN` as it's being committed, but it will not log the change to the event history.

The original Zino features its own function to change the state of an event, and this function will log changes to the history, among other things.  It also takes an optional `user` argument, which will cause it to log the name of the user who initiated the state change (this is only ever used by the `SETSTATE` API command).

Something similar is likely needed on the `Event` class.

lunkwill42:
This reworks the workflow for creating new events or changing existing ones, to be more inline with the *commit* workflow introduced by #136.

The new workflow is mostly concerned with changes to the `Events` index class:

* The `create()` method will not place a created `Event` object in the index.  The returned object **must** be explicitly committed to the index afterwards, using the `Events.commit()` method.
* The `get_or_create()` method follows the same logic, but also: When an existing `Event` is found and returned, it performs a `checkout()` on this event.  This means the caller is always expected to run `Events.commit()` for any `Event` object returned by this call in order to make sure any event changes are made "official".
* The `get_or_create()` method no longer returns a tuple where the second value is a `created` boolean.  Rather, a new `Event` can be now be distinguished by the fact that *it's `id` attribute is `None`.  A new `Event` will have its `id` attribute assigned by the `Events` index on the first call to `commit()` it.

Additionally, this fixes #137 by adding a `set_state()` method to the `Event` class, which encapsulates the typical operations of changing an event's state and logging this change to the event's history, including the username of the user who made the change.  It also removes all code instances that explicitly set the event state to `OPEN` when it discovered it had created a new event.

* The `Events.commit()` method is updated to use the new `set_state()` when embryonic events are commited.  It is also updated to take an optional `user` (a username) argument, which can be used by the API to ensure changes made by specific users are logged with that user's username.  If omitted, it will default to use `monitor` username, which is what the original Tcl code does.

### DIFF
--- a/tests/events_test.py
+++ b/tests/events_test.py
@@ -15,17 +15,18 @@ class TestEvents:
         events = Events()
         event = events.create_event("foobar", None, ReachabilityEvent)
 
-        assert event.id > 0
+        assert isinstance(event, ReachabilityEvent)
 
-    def test_event_registry_should_contain_one_event_when_first_event_is_created(self):
+    def test_event_registry_should_not_contain_newly_created_event(self):
         events = Events()
         events.create_event("foobar", None, ReachabilityEvent)
 
-        assert len(events) == 1
+        assert len(events) == 0
 
     def test_adding_two_identical_events_should_raise(self):
         events = Events()
-        events.create_event("foobar", None, ReachabilityEvent)
+        event = events.create_event("foobar", None, ReachabilityEvent)
+        events.commit(event)
 
         with pytest.raises(EventExistsError):
             events.create_event("foobar", None, ReachabilityEvent)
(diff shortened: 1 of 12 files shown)


johannaengland (approved):

tests/events_test.py
```diff
@@ -15,52 +15,57 @@ def test_create_event_should_return_event(self):
         events = Events()
         event = events.create_event("foobar", None, ReachabilityEvent)
 
-        assert event.id > 0
+        assert isinstance(event, ReachabilityEvent)
 
-    def test_event_registry_should_contain_one_event_when_first_event_is_created(self):
+    def test_event_registry_should_not_contain_newly_created_event(self):
```
johannaengland, line 20: This by itself doesn't make it clear for me why it should not be in the event registry, maybe something like this is clearer:

```suggestion
    def test_event_registry_should_not_contain_uncommited_event(self):
```

johannaengland (approved):
Just a little comment about naming of tests

tests/events_test.py
lunkwill42: Much better - fixed now. Thanks!

hmpf (approved):

PR #139, opened 2024-01-10
Adding tests for bgp state monitor 

johannaengland:
Adding tests for external reset. Adding assertions about the created event to existing tests. Adding a missing condition. Make snmprec files more similar to each other.


stveit (changes requested):
nitpickery

tests/snmp_fixtures/general-bgp-external-reset.snmprec
```diff
@@ -0,0 +1,9 @@
+1.3.6.1.2.1.1.3.0|67|1000000
+1.3.6.1.2.1.15.2.0|2|10
+1.3.6.1.2.1.15.3.1.2.10.0.0.1|2|6
+1.3.6.1.2.1.15.3.1.3.10.0.0.1|2|2
+1.3.6.1.2.1.15.3.1.5.10.0.0.1|64x|0a000002
+1.3.6.1.2.1.15.3.1.7.10.0.0.1|64x|0a000002
```
stveit, line 6: Should these have the same value? or does it not matter?

tests/tasks/test_bgpstatemonitortask.py
```diff
@@ -111,18 +292,103 @@ async def test_peer_admin_status_changing_to_stop_while_peer_state_is_not_establ
...
+    async def test_admin_up_cisco_creates_event(self, snmpsim, snmp_test_port):
+        device = PollDevice(
+            name=DEVICE_NAME,
+            address=DEVICE_ADDRESS,
+            community="cisco-bgp-admin-up",
+            port=snmp_test_port,
+        )
+        state = ZinoState()
+        task = BgpStateMonitorTask(device, state)
+        peer_address = IPv4Address("10.0.0.1")
```
stveit, line 346: this is another candidate for "normalization". Cisco used `10.0.0.1` for `peer_adress` while juniper and general uses `10.0.0,2`

tests/tasks/test_bgpstatemonitortask.py
```diff
@@ -53,6 +53,181 @@ async def test_task_logs_missing_information(self, snmpsim, snmp_test_port, capl
...
+        event = task.state.events.get(device_name=task.device.name, port=peer_address, event_class=BGPEvent)
+        assert event
+        assert event.admin_status == "running"
+        assert event.operational_state == "established"
+        assert event.remote_address == peer_address
+        assert event.remote_as == 20
+        assert event.peer_uptime == 250
+
+    @pytest.mark.asyncio
+    async def test_session_up_general_creates_event(self, snmpsim, snmp_test_port):
```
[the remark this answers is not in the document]
johannaengland: That's true. I updated the three session up tests. For the rest in production these events will mostly be updated and rarely created, but in the other tests they are, so I'm keeping the rest of the names as is.

tests/tasks/test_bgpstatemonitortask.py
```diff
@@ -53,6 +53,181 @@ async def test_task_logs_missing_information(self, snmpsim, snmp_test_port, capl
...
+        event.remote_as = 20
+        event.peer_uptime = 0
+
+        await task.run()
+
+        # check if state has been updated to reflect state defined in .snmprec
+        assert task.device_state.bgp_peer_admin_states[peer_address] == "start"
+        assert task.device_state.bgp_peer_oper_states[peer_address] == "established"
+        # check that the correct event has been created
+        event = task.state.events.get(device_name=task.device.name, port=peer_address, event_class=BGPEvent)
```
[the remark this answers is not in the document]
johannaengland: Good catch!

tests/snmp_fixtures/general-bgp-external-reset.snmprec
johannaengland: For these tests it does not matter

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -290,21 +291,18 @@ def _update_single_bgp_entry(self, oid, row: dict[str, Any], local_as: str, upti
...
         if data.peer_state == "established":
-            handled = False
             bgp_peer_up_time = self.device_state.bgp_peer_up_times.get(data.peer_remote_address, None)
-            if (
-                bgp_peer_up_time
-                and uptime >= bgp_peer_up_time
-                and bgp_peer_up_time > data.peer_fsm_established_time
-                and local_as is not data.peer_remote_as
-            ):
+            if bgp_peer_up_time and uptime >= bgp_peer_up_time and bgp_peer_up_time > data.peer_fsm_established_time:
```
stveit, line 302:
```suggestion
            if bgp_peer_up_time and uptime >= bgp_peer_up_time > data.peer_fsm_established_time:
```
I think its valid to simplify like this. I'll let you decide if its a good choice, whatever gives best readability should be used

tests/tasks/test_bgpstatemonitortask.py
```diff
@@ -54,9 +54,185 @@ async def test_task_logs_missing_information(self, snmpsim, snmp_test_port, capl
...
-    ):
+    async def test_external_reset_general_creates_event(self, snmpsim, snmp_test_port):
+        device = PollDevice(
+            name=DEVICE_NAME,
+            address=DEVICE_ADDRESS,
+            community="general-bgp-external-reset",
+            port=snmp_test_port,
+        )
+        state = ZinoState()
+        task = BgpStateMonitorTask(device, state)
```
stveit, line 65: We did discuss a bit about reemoving redundancy from the tests with conclusion that the tests being very transparent in what is being done being better than fewer lines. But I think its fine to move these lines to its own Fixture. The only thing from these being used later on in the test is the `task` object, so making a `general_external_reset_task` fixture (and similar for the other snmprec files) would make the tests easier to read by removing complexity, without removing any relevant context. the `DEVICE_NAME`, `DEVICE_ADDRESS` etc being used to define the PollDevice is largely irrelevant for these tests.

The fixture name would still make it clear what snmprec file is being used, and that we are operating on a `BgpMonitorTask` object.

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -12,13 +12,14 @@
...
     BgpOperStatus,
     BgpStyle,
-    EventState,
     IPAddress,
 )
 from zino.tasks.task import Task
 
 _logger = logging.getLogger(__name__)
 
+TIME_BEFORE_OPER_DOWN_ALERT = 600
```
stveit, line 21: nice

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -290,21 +291,18 @@ def _update_single_bgp_entry(self, oid, row: dict[str, Any], local_as: str, upti
         if data.peer_remote_address == "32.1.7.0":
             return
 
+        # Internal bgp sessions are not observed
+        if local_as == data.peer_remote_as:
+            return
```
stveit, line 296: Early return here instead of nesting everything below it is a nice touch

stveit (commented):
The "commit" feature being added in this PR made it a bit difficult to review the BGP specific changes since its so much going on. I wont comment on the commit/checkout features, ill focus on BGP stuff and assume the other stuff works fine.

I dont see anything thats obviously wrong. 

image
 Couldnt make inline comment since this is not part of this PR, but using the `ipaddress.ip_address` function can be used to automatically detect ipv4/ipv6. can be done in a different PR tho, this one already has a lot going on.

There are some lines missing tests, these seem to be outliers for when things go wrong. Might be worth making some additional test to cover these as well.

I'll leave this review as a comment, you can decide yourself if any of these suggestions are useful. The commit story is obviously a bit fucked and not ready to be merged so I wont approve either :P

tests/tasks/test_bgpstatemonitortask.py
johannaengland: Thanks for giving me that push and sending me down the pytest rabbit hole! I was able to strongly simplify the tests!

PR #143, opened 2024-01-12
Make event commit update event index on new and old events

johannaengland:
Before this updating only happened on committing a new event.


lunkwill42 (approved):
A big, resounding YES! :)

PR #145, opened 2024-01-12
Collect and map router IP addresses

### LINKED ISSUE #141: Make task to collect router IP addresses and make two-way maps in Zino state
Zino 1 maintains a two-way mapping between router names and IP addresses.  It seems like this is mostly used to be able to map the source address of incoming SNMP traps to known routers from `polldevs.cf`.

These maps are mostly built from `IP-MIB::ipAdEntAddr`, which means only IPv4 is supported, for now.

lunkwill42:
This introduces a new task to collect router IP address and add forward and reverse maps to the Zino server state, as the original Zino code does.

This includes a hardcoded list of ignored IP addresses and IP address patterns, as does Zino 1.

Closes #141


johannaengland (approved):
Very good that you immediately found out that the ignore mechanism works :grin: 

Only two potential improvements:
1. Adding a test for `IP-Mib` in `snmp_test.TestMibResolver`
2. Renaming `ipadentaddr.snmprec` to `ip-ad-ent-addr.snmprec` to make it more human readable

lunkwill42:
> Only two potential improvements:

Yeah, I'll buy those! :) Committed.

johannaengland (approved):

PR #146, opened 2024-01-16
Use pytest parameters to avoid duplicate tests for BFD

stveit:
title


johannaengland (approved):
Looks good, the only thing that confuses me is the grouping of the tests - why have the class TestJuniper for two tests and then the rest of the tests outside of a class?

stveit:
> Looks good, the only thing that confuses me is the grouping of the tests - why have the class TestJuniper for two tests and then the rest of the tests outside of a class?

I dont remember why but those two test juniper specific functions. Should maybe be similar tests for cisco, but I see that as outside the scope of this PR

PR #148, opened 2024-01-17
Initial implementation of COMMUNITY API command

### LINKED ISSUE #112: Implement the `COMMUNITY` API command
The `COMMUNITY` API command takes a router name as its input and responds with the configured SNMP community for the given router.

Example session from Zino 1:

```
COMMUNITY example-gw
201 public
```

The original implementation doesn't seem to have a fault handler, meaning it gives no response at all if the named router does not exist.  IMHO Zino2 should give a proper 500 error response if the name is unknown.

johannaengland:
Closes #112.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -301,6 +301,16 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
 
         return self._respond_ok()
 
+    @requires_authentication
+    async def do_community(self, router_name: str):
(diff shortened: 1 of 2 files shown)


lunkwill42 (approved):
This is correct :)

I think a future consideration should be to also infuse the protocol object with configuration state, just as we infuse it with running server state - the only part about this PR I don't like is the import of the global `zino.state.polldevs` structure.   I can live with it for now, but we should change it down the road.

PR #149, opened 2024-01-18
Add notification protocol

### LINKED ISSUE #147: Implement the notification mechanism and the `NTIE` API command
Zino 1's notification mechanism is loosely described thus:

- A client that wishes to receive a continuous stream of notifications about Zino event changes must open a new TCP connection to port **8002** (the main API protocol is on port 8001).
- Once connected to port 8002, the client will receive a nonce value (a random hexstring) from Zino.
- The client must now run the `NTIE` command on its regular connection to port 8001, supplying the received nonce value as the command argument.  This serves to identify the client's connection to the notification channel on port 8002.
- Zino now commences to emit event change notifications on the notification channel.  The notification protocol is also a simple text-based protocol, where each notification is a line of text followed by a newline.
- If the client closes the connection on port 8001, Zino also closes the associated notification channel.

lunkwill42:
This adds a notification protocol implementation, as well as refactoring how the two protocol servers are set up when running Zino.

API protocol and notify protocol instances need to be able to look up each other's existence in order for notification channels to be properly authenticated, so this introduces a new `ZinoServer` class that takes the responsibility of setting up both asyncio servers and keeping track of which client sessions exist for both.

Each protocol instance now also takes a reference to the controlling `ZinoServer` object as an __init__ argument, so as to be able to register themselves, and to be able to look each other up when notification channels are being set up and associated.

`zino.zino.main` is refactored to delegate responsibility for setting up the servers to `ZinoServer`.

I potentially recommend reviewing this commit-by-commit.

Closes #147

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -10,12 +10,15 @@ import re
 import textwrap
 from functools import wraps
 from pathlib import Path
-from typing import Callable, List, Optional, Union
+from typing import TYPE_CHECKING, Callable, List, Optional, Union
 
 from zino import version
 from zino.api import auth
 from zino.state import ZinoState
 from zino.statemodels import Event, EventState
+
+if TYPE_CHECKING:
+    from zino.api.server import ZinoServer
 
 _logger = logging.getLogger(__name__)
 
@@ -29,13 +32,19 @@ def requires_authentication(func: Callable) -> Callable:
 class Zino1BaseServerProtocol(asyncio.Protocol):
     """Base implementation of the Zino 1 protocol, with a basic command dispatcher for subclasses to utilize."""
 
-    def __init__(self, state: Optional[ZinoState] = None, secrets_file: Optional[Union[Path, str]] = "secrets"):
+    def __init__(
+        self,
+        server: Optional["ZinoServer"] = None,
+        state: Optional[ZinoState] = None,
+        secrets_file: Optional[Union[Path, str]] = "secrets",
+    ):
         """Initializes a protocol instance.
 
         :param state: An optional reference to a running Zino state that this server should be based on.  If omitted,
                       this protocol will create and work on an empty state object.
         :param secrets_file: An optional alternative path to the file containing users and their secrets.
         """
+        self.server = server
         self.transport: Optional[asyncio.Transport] = None
         self._authenticated_as: Optional[str] = None
         self._current_task: asyncio.Task = None
@@ -65,9 +74,16 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
 
     def connection_made(self, transport: asyncio.Transport):
         self.transport = transport
-        _logger.debug("New server connection from %s", self.peer_name)
+        _logger.info("New server connection from %s", self.peer_name)
+        if self.server:
+            self.server.active_clients.add(self)
         self._authentication_challenge = auth.get_challenge()
         self._respond_ok(f"{self._authentication_challenge} Hello, there")
+
+    def connection_lost(self, exc: Optional[Exception]) -> None:
+        _logger.info("Client disconnected: %s", self.peer_name)
+        if self.server:
+            self.server.active_clients.remove(self)
 
     def data_received(self, data):
         try:
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -59,7 +59,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         self._secrets_file = secrets_file
 
     @property
-    def peer_name(self) -> str:
+    def peer_name(self) -> Optional[str]:
         return self.transport.get_extra_info("peername") if self.transport else None
 
     @property
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -337,7 +337,7 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
         """Implements the NTIE command that ties together this session with a notification channel."""
         try:
             channel = self.server.notification_channels[nonce]
-        except KeyError:
+        except (AttributeError, KeyError):
             return self._respond_error("Could not find your notify socket")
 
         self.notification_channel = channel
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING, Callable, List, Optional, Union
 
 from zino import version
 from zino.api import auth
+from zino.api.notify import Zino1NotificationProtocol
 from zino.state import ZinoState
 from zino.statemodels import Event, EventState
 
@@ -46,6 +47,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         """
         self.server = server
         self.transport: Optional[asyncio.Transport] = None
+        self.notification_channel: Optional[Zino1NotificationProtocol] = None
         self._authenticated_as: Optional[str] = None
         self._current_task: asyncio.Task = None
         self._multiline_future: asyncio.Future = None
@@ -84,6 +86,8 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         _logger.info("Client disconnected: %s", self.peer_name)
         if self.server:
             self.server.active_clients.remove(self)
+        if self.notification_channel:
+            self.notification_channel.goodbye()
 
     def data_received(self, data):
         try:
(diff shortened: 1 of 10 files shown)


src/zino/api/legacy.py
```diff
@@ -29,14 +33,21 @@ def requires_authentication(func: Callable) -> Callable:
 class Zino1BaseServerProtocol(asyncio.Protocol):
     """Base implementation of the Zino 1 protocol, with a basic command dispatcher for subclasses to utilize."""
 
-    def __init__(self, state: Optional[ZinoState] = None, secrets_file: Optional[Union[Path, str]] = "secrets"):
+    def __init__(
+        self,
+        server: Optional["ZinoServer"] = None,
```
hmpf, line 38: Are the quotes needed here?

src/zino/api/legacy.py
```diff
@@ -29,14 +33,21 @@ def requires_authentication(func: Callable) -> Callable:
...
+        state: Optional[ZinoState] = None,
+        secrets_file: Optional[Union[Path, str]] = "secrets",
+    ):
         """Initializes a protocol instance.
 
         :param state: An optional reference to a running Zino state that this server should be based on.  If omitted,
                       this protocol will create and work on an empty state object.
         :param secrets_file: An optional alternative path to the file containing users and their secrets.
         """
+        self.server = server
```
hmpf, line 48: Add ":param server: An optional instance of ZinoServer" above.

src/zino/api/legacy.py
```diff
@@ -311,6 +331,20 @@ async def do_community(self, router_name: str):
         else:
             self._respond_error("router unknown")
 
+    @requires_authentication
+    async def do_ntie(self, nonce: str):
+        """Implements the NTIE command that ties together this session with a notification channel."""
+        try:
+            channel = self.server.notification_channels[nonce]
```
hmpf, line 338: If self.server is None, you'll never get as far as the KeyError.

src/zino/api/server.py
```diff
@@ -0,0 +1,48 @@
...
+
+class ZinoServer:
+    """Represents the two asyncio servers that work in tandem to implement the Zino 1 legacy API:
+
+    Port 8001 is the text-based command interface.
+    Port 8002 is the text-based notification interface.
+    """
+
+    API_PORT = 8001
+    NOTIFY_PORT = 8002
```
hmpf, line 21: Are these also hardcoded in Zino 1? Curitz, and thus howitz, can be configured to work with other ports than this.

tests/api/notify_test.py
```diff
@@ -0,0 +1,139 @@
...
+from zino.state import ZinoState
+from zino.statemodels import EventState, ReachabilityEvent
+from zino.time import now
+
+
+class TestZino1NotificationProtocol:
+    def test_init_should_succeed(self):
+        assert Zino1NotificationProtocol()
+
+    def test_when_unconnected_then_peer_name_should_be_none(self):
```
hmpf, line 17: "test_when_not_connected.." is better, surely.

hmpf (changes requested):
Many questions, one change.

src/zino/api/legacy.py
hmpf: .. github missed the line. You have `Optional["ZinoServer"]` with quotes.

tests/api/notify_test.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.fixture
+def fake_event() -> ReachabilityEvent:
+    return ReachabilityEvent(id=42, router="example-gw.example.org", state=EventState.OPEN)
+
+
+@pytest.fixture
+def changed_fake_event(fake_event) -> ReachabilityEvent:
+    copy = fake_event.model_copy(deep=True)
+    copy.add_history("this fake event has been changed")
+    return copy
```
johannaengland, line 139: Do we actually need to make a fixture out of it if it is only used once?

johannaengland (changes requested):

src/zino/api/notify.py
```diff
@@ -0,0 +1,57 @@
+"""Notification channel implementation for Zino 2.0.
+
+Notification channels are currently part of the legacy API from the Tcl-based Zino 1.0, and is a simple text-based,
```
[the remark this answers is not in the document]
hmpf: How about "Notification channels are currently part of the legacy API from the Tcl-based Zino 1.0. They are a simple text-based,"

src/zino/api/legacy.py
lunkwill42: > Are the quotes needed here?

Apparently, since the symbol is only imported `if typing.TYPE_CHECKING`.  Tests fail if I leave it unquoted.

lunkwill42: > .. github missed the line. You have `Optional["ZinoServer"]` with quotes.

Huh?

src/zino/api/server.py
lunkwill42: Indeed, they are.

tests/api/notify_test.py
lunkwill42: Indeed, but the name mirrors the same test for the base server protocol in `legacy_test`, hence the name.  If I change this, both should be renamed.

src/zino/api/legacy.py
```diff
@@ -65,10 +74,17 @@ def user(self, user_name: str):
...
     def connection_made(self, transport: asyncio.Transport):
         self.transport = transport
-        _logger.debug("New server connection from %s", self.peer_name)
+        _logger.info("New server connection from %s", self.peer_name)
+        if self.server:
+            self.server.active_clients.add(self)
         self._authentication_challenge = auth.get_challenge()
         self._respond_ok(f"{self._authentication_challenge} Hello, there")
 
+    def connection_lost(self, exc: Optional[Exception]) -> None:
```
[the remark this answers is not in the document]
lunkwill42: Not by our code, but this is the signature of `BaseProtocol.connection_lost()`, which we are overriding, so we need to keep it like this.

Ref.:
https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseProtocol.connection_lost

tests/api/notify_test.py
```diff
@@ -0,0 +1,50 @@
...
+from zino.state import ZinoState
+
+
+class TestZino1NotificationProtocol:
+    def test_init_should_succeed(self):
+        assert Zino1NotificationProtocol()
+
+    def test_when_unconnected_then_peer_name_should_be_none(self):
+        protocol = Zino1NotificationProtocol()
+        assert not protocol.peer_name
```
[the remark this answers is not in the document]
lunkwill42: Also, the signature and test is copied from the server protocol tests, so they should both change in the same way.

tests/api/notify_test.py
lunkwill42: No, but I like having it around, as the explicit naming and separation of concerns appeal to me.

src/zino/api/legacy.py
lunkwill42: fixed

tests/api/notify_test.py
lunkwill42: fixed

src/zino/api/notify.py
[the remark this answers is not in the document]
lunkwill42: @johannaengland's suggestion is good. @hmpf's is even better :)  Fixed!

src/zino/api/notify.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Zino1NotificationProtocol(asyncio.Protocol):
+    """Basic implementation of the Zino 1 notification protocol"""
+
+    def __init__(self, server: Optional["ZinoServer"] = None, state: Optional[ZinoState] = None):
+        """Initializes a protocol instance.
+
+        :param state: An optional reference to a running Zino state that this server should be based on.  If omitted,
+                      this protocol will create and work on an empty state object.
+        """
```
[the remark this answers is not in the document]
lunkwill42: fixed!

tests/api/notify_test.py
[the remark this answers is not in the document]
lunkwill42: fixed for both versions of the tests and `peer_name` implementations (even if it gets slightly out of scope for this PR)

src/zino/api/legacy.py
hmpf: Github or my browser or a combo sometimes hides lines.

hmpf (approved):

src/zino/api/legacy.py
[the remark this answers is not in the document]
johannaengland: Ah, I see

johannaengland (approved):

PR #150, opened 2024-01-19
Include new and old event references in event observer functions

lunkwill42:
This facilitates a change in the event observer interface that is necessary for the implementation of #149 to move forward.

An event observer function needs more than just the ID of a changed event to be able to do many useful things, especially if that event observer needs to inspect which attributes of an event changed and to queue notifications as a result of this.

This PR changes the event observer interface to include the full old and new event objects as arguments to observer functions (where the old object is optional, since there isn't one the first time an event is committed).

### DIFF
--- a/tests/zino_test.py
+++ b/tests/zino_test.py
@@ -36,17 +36,19 @@ class TestZinoRescheduleDumpStateOnCommit:
     def test_when_more_than_10_seconds_remains_until_next_dump_it_should_reschedule(self):
         scheduler = get_scheduler()
         mock_job = Mock(next_run_time=now() + timedelta(minutes=5))
+        mock_event = Mock(id=42)
 
         with patch.object(scheduler, "get_job") as get_job:
             get_job.return_value = mock_job
-            zino.reschedule_dump_state_on_commit(42)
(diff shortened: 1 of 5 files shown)


tests/zino_test.py
```diff
@@ -36,10 +36,11 @@ class TestZinoRescheduleDumpStateOnCommit:
     def test_when_more_than_10_seconds_remains_until_next_dump_it_should_reschedule(self):
         scheduler = get_scheduler()
         mock_job = Mock(next_run_time=now() + timedelta(minutes=5))
+        mock_event = Mock(id=42)
 
         with patch.object(scheduler, "get_job") as get_job:
             get_job.return_value = mock_job
-            zino.reschedule_dump_state_on_commit(42)
+            zino.reschedule_dump_state_on_commit(mock_event)
```
johannaengland, line 43: `test_when_less_than_10_seconds_remains_until_next_dump_it_should_not_reschedule` should also be amended like this

johannaengland (changes requested):

johannaengland (approved):

PR #153, opened 2024-01-22
Factor out get uptime

johannaengland:


src/zino/statemodels.py
```diff
@@ -108,9 +108,9 @@ def is_juniper(self):
     def set_boot_time_from_uptime(self, uptime: int):
         """Calculates and sets the device boot time from a current uptime value.
 
-        :param uptime: An uptime value in 100ths of a second
+        :param uptime: An uptime value in seconds
         """
-        self.boot_time = now() - datetime.timedelta(seconds=uptime / 100)
+        self.boot_time = now() - datetime.timedelta(seconds=uptime)
```
lunkwill42, line 113: This moves the onus on the collector to convert from the centiseconds value collected from `sysUpTime`.  Why?

src/zino/tasks/task.py
```diff
@@ -16,6 +16,12 @@ def __init__(self, device: PollDevice, state: ZinoState):
     async def run(self):
         """Runs job asynchronously"""
 
+    async def _get_uptime(self) -> int:
+        """Polls and returns the device sysuptime value rounded to seconds"""
+        response = await self.snmp.get("SNMPv2-MIB", "sysUpTime", 0)
+        uptime = response.value
+        return round(uptime / 100)
```
lunkwill42, line 23: You just reduced the precision of the uptime value by rounding it.  Why?

lunkwill42 (commented):
I would have liked some discussion on these changes before they were merged :P

PR #154, opened 2024-01-22
Return sysuptime in 100ths of a second

johannaengland:
Reverse all associated changes from #153.


hmpf (approved):

lunkwill42 (approved):
Much better.  We keep the precision we get from the source, and only reduce precision in the presentation layer.

PR #159, opened 2024-01-23
Improve packaging/testing

hmpf:
We should perhaps also start using a requirements.lock file.


hmpf:
I wish we could use tox 4, but tox 4 fails to install the dependencies of zino. If one then uses a requirtements.lock-file to explicitly install dependencies, some tests fail.

lunkwill42 (approved):

PR #160, opened 2024-01-23
Add short installation/configuration section to `README.md`

### LINKED ISSUE #156: Add a small installation guide to `README.md`

lunkwill42:
This documents how to install both from PyPI and from source.  **NOTE** however, Zino is not yet published on PyPI, but the guide should work once it's published.

Closes #156

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,15 @@
 This is the modern Python re-implementation of the battle-proven Zino network
 state monitor, first implemented in Tcl/Scotty at Uninett in the 1990s.
 
-This is still a work in progress, and is not yet a fully functional
-replacement for the Tcl-based Zino.
+This is still a work in progress, and is not yet a fully functional replacement
+for the original Tcl-based Zino.  An incomplete list of features that have yet
+to be ported:
+
+- SNMP trap reception
+  - Hence, no port flapping status
+- Planned maintenance (the `PM` family of API commands)
+- Router and single-port poll triggering from the API (the `POLLRTR` and
+  `POLLINTF` API commands)
 
 Development of Zino 2.0 is fully sponsored by [NORDUnet](https://nordu.net/),
 on behalf of the nordic NRENs.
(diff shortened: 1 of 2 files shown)


polldevs.cf.example
```diff
@@ -0,0 +1,34 @@
...
+#
+# An empty line signifies the start of a new configuration block, i.e. a set of
+# defaults, or a new, distinct router.  Router names can be arbitrary, but must
+# be unique across all of Zino.  Comments begin with the # character.
+
+# Default values that will apply to all router entries unless overridden:
+default interval: 5
+default community: public
+default domain: example.org
+default hcounters: yes
```
johannaengland, line 11: Would be nice to have more information about those default values (specifically interval and hcounters)

README.md
```diff
@@ -12,7 +12,74 @@ replacement for the Tcl-based Zino.
...
+git clone https://github.com/Uninett/zino.git
+cd zino
+pip install .
+```
+
+### Running Zino for the first time
+
+You should be able to run Zino (although you will need to configure a list of
+devices to poll in `polldevs.cf` first) from the command line as long as its
+virtual environment is activated:
```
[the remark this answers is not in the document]
johannaengland: Or in some other way point towards the section about `polldevs.cf`

johannaengland (changes requested):

polldevs.cf.example
hmpf: What is the time unit for interval: seconds? minutes?

README.md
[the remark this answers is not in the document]
lunkwill42: Then the whole example needs to change as well, since it demonstrates Zino's error message when there is no `polldevs.cf` present.

polldevs.cf.example
lunkwill42: minutes, yeah. It says so further down in the file, when `interval` is specified again, but I've added a comment to the example.

lunkwill42:
Since we added the small configuration section, I realized we should also document how to configure user accounts for the API, so I've added yet another section for that.  Feel free to have another look :)

johannaengland (approved):

hmpf (approved):

PR #161, opened 2024-01-24
Replace overloaded `port` concept for indexing of events

lunkwill42:
The original Zino 1 Tcl code used `port` as one of three components used to uniquely index events.  However, it greatly overloaded the meaning of `port`.  

For reachability events, `port` would be an empty value, since the event would pertain to the router itself, and not to a sub-component of the router.  For actual port state events, it would be the `ifindex` value of the port in question.  For BGP events, it would be the IP address of the BGP peer, and so forth.

We took this concept and ran with it in the initial stages of porting the code to Python, but the concept leaked into the attributes of the `Event` base class, causing all event types to have a `port` attribute, which made no sense.  This became rather apparent when writing a converter for Zino 1 state to Zino 2 state.

So, this is the time to get rid of this concept and be more explicit about an index value being an index value, and not a port name or identifier.

Each event type, with the notable exception of reachability events, has some attribute that uniquely identifies it among events of the same type for the same router.  This attribute is named differently for the different event types.  Since some of the existing event types relied on its inherited `port` attribute for this, this PR adds new attributes to these event types (to match the names given in the Zino 1 API), and replaces the concept of `port` for indexing with a generic `subindex` term.

The pre-existing `PortOrIPAddress` compound type annotation was mean to represent these index values, but was badly named (as it encompassed more than just ports and IP addresses).  This was renamed and repurposed into a `SubIndex` compound type, which can now be extended if new event types are ever added.

The PR also fixes a few small bugs and issues with how event attributes were set (or not set) by the various implemented tasks.

### DIFF
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -65,6 +65,7 @@ class JuniperAlarmTask(Task):
         )
 
         old_alarm_count = alarm_event.alarm_count
+        alarm_event.alarm_type = color
         alarm_event.alarm_count = alarm_count
         alarm_event.add_log(f"{self.device.name} {color} alarms went from {old_alarm_count} to {alarm_count}")
         self.state.events.commit(alarm_event)
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -60,7 +60,7 @@ class JuniperAlarmTask(Task):
     def create_alarm_event(self, color: AlarmType, alarm_count: int):
         alarm_event = self.state.events.get_or_create_event(
             device_name=self.device.name,
-            port=color,
+            subindex=color,
             event_class=AlarmEvent,
         )
 
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -1,7 +1,6 @@
 import logging
-from typing import Literal
 
-from zino.statemodels import AlarmEvent
+from zino.statemodels import AlarmEvent, AlarmType
 from zino.tasks.task import Task
 
 _logger = logging.getLogger(__name__)
@@ -58,7 +57,7 @@ class JuniperAlarmTask(Task):
 
         return yellow_alarm_count, red_alarm_count
 
-    def create_alarm_event(self, color: Literal["yellow", "red"], alarm_count: int):
+    def create_alarm_event(self, color: AlarmType, alarm_count: int):
         alarm_event = self.state.events.get_or_create_event(
             device_name=self.device.name,
             port=color,
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -68,4 +68,6 @@ class JuniperAlarmTask(Task):
         alarm_event.alarm_type = color
         alarm_event.alarm_count = alarm_count
         alarm_event.add_log(f"{self.device.name} {color} alarms went from {old_alarm_count} to {alarm_count}")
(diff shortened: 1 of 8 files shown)


src/zino/tasks/juniperalarmtask.py
```diff
@@ -58,14 +57,17 @@ async def _get_juniper_alarms(self):
 
         return yellow_alarm_count, red_alarm_count
 
-    def create_alarm_event(self, color: Literal["yellow", "red"], alarm_count: int):
+    def create_alarm_event(self, color: AlarmType, alarm_count: int):
```
hmpf, line 60: SOOO MUCH BETTER

src/zino/tasks/reachabletask.py
```diff
@@ -28,6 +28,8 @@ async def run(self):
             if event.reachability != ReachabilityState.NORESPONSE:
                 event.reachability = ReachabilityState.NORESPONSE
                 event.add_log(f"{self.device.name} no-response")
+            event.polladdr = self.device.address
+            event.priority = self.device.priority
```
hmpf, line 32: Why are polladdr and priority also added to every event?

tests/tasks/test_juniperalarmtask.py
```diff
@@ -72,8 +72,8 @@ async def test_task_creates_both_alarm_events_on_both_counts_changed(self, junip
 
         await task.run()
 
-        yellow_event = task.state.events.get(device_name=task.device.name, port="yellow", event_class=AlarmEvent)
-        red_event = task.state.events.get(device_name=task.device.name, port="red", event_class=AlarmEvent)
+        yellow_event = task.state.events.get(device_name=task.device.name, subindex="yellow", event_class=AlarmEvent)
+        red_event = task.state.events.get(device_name=task.device.name, subindex="red", event_class=AlarmEvent)
```
hmpf, line 76: Not literal of AlarmType? Might be more trouble than it is worth.

src/zino/tasks/juniperalarmtask.py
lunkwill42: It was already there, it was just a matter of actually using it.

src/zino/tasks/reachabletask.py
lunkwill42: Because Zino 1 does it, and it was neglected from several of the task plugins.

It could be argued that it doesn't belong in this particular PR, but I discovered it while comparing Zino 1 events to Zino 2 events during this PR, so I threw it in there for good measure :-D

tests/tasks/test_juniperalarmtask.py
lunkwill42: Not quite sure what you're getting at. `"red"` and `"yellow"` are valid `AlarmType` values:
```python
AlarmType = Literal["yellow", "red"]
```

johannaengland (approved):
Very very happy with this PR, makes things a lot cleaner. 

The only small request that I have is that the docstrings for `get_or_create_event` and `create_event` should be adjusted to indicate that event though a subindex is given it will not automatically be set for the field that represents the subindex.

tests/tasks/test_juniperalarmtask.py
hmpf: I was thinking that AlarmType was an enum. Mypy can be a bit stupid about enums, that's why I wondered if t was worth it.

hmpf (approved):

lunkwill42:
> The only small request that I have is that the docstrings for `get_or_create_event` and `create_event` should be adjusted to indicate that event though a subindex is given it will not automatically be set for the field that represents the subindex.

Pushed one more commit to this effect, then.

PR #164, opened 2024-01-25
 Implement line buffering in data_received() 

lunkwill42:
While manually testing the protocol implementation using simple tools like telnet or netcat, `data_received()` will usually receive full text lines from the transport, as these tools typically send a packet once Enter is pressed.

However, `Protocol.data_received()` will in fact receive any old piece of binary data, which can potentially come in arbitrarily sized packets, depending on the client.  As demonstrated by the legacy cuRitz client, which will for example dispatch the text data associated with a `ADDHIST` command as one large, multiline packet: This causes the naive implementation of `data_received()` to fail and hang forever, as the multiline input mode will be waiting for a single packet containing `b'.\r\n'` before it's satisfied that input has ended.

This reworks `data_received()` to employ a simple bytearray buffer of incoming data, and moves most of the message parsing logic to a new method `message_received()`.  The latter will only receive single, full lines of protocol data as string objects, with newlines and carriage returns stripped.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -51,6 +51,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         self.notification_channel: Optional[Zino1NotificationProtocol] = None
         self._authenticated_as: Optional[str] = None
         self._current_task: asyncio.Task = None
+        self._input_buffer = bytearray()
         self._multiline_future: asyncio.Future = None
         self._multiline_buffer: List[str] = []
         self._authentication_challenge: Optional[str] = None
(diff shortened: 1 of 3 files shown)


src/zino/api/legacy.py
```diff
@@ -91,13 +92,18 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
...
-        try:
-            message = data.decode().rstrip("\r\n")
-        except UnicodeDecodeError:
-            _logger.error("Received garbage server input from %s: %r", self.peer_name, data)
-            self.transport.close()
-            return
-        _logger.debug("Data received from %s: %r", self.peer_name, message)
+        self._input_buffer.extend(data)
+        while b"\n" in self._input_buffer:
+            line, self._input_buffer = self._input_buffer.split(b"\n", 1)
```
hmpf, line 97: Telnet sends `\r\n` but curitz may send `\n`.

tests/api/legacy_test.py
```diff
@@ -90,6 +90,18 @@ async def test_read_multiline_should_return_data_as_future(self):
...
         assert data == ["line one", "line two"]
 
+    @pytest.mark.timeout(5)
+    @pytest.mark.asyncio
+    async def test_data_received_should_break_down_multiline_input_packets(self):
+        protocol = Zino1BaseServerProtocol()
+        fake_transport = Mock()
+        protocol.connection_made(fake_transport)
+        future = protocol._read_multiline()
+        protocol.data_received(b"line one\r\nline two\r\n.\r\n")
```
hmpf, line 100: You're not testing for just `\n`.

src/zino/api/legacy.py
lunkwill42: Indeed, which is why the code splits on `\n` and strips `\r`, which should support both cases.

tests/api/legacy_test.py
lunkwill42: Fixed

hmpf (approved):
Postel would be happy :)

Issue #10, opened 2023-05-08
Research how SonarQube's scans can be augmented with linter reports from `ruff`.


johannaengland:
It might also be interesting at this point to research how SonarQube's scans can be augmented with linter reports from `ruff` (which this project uses).

I did pick up somewhere that SQ is supposed to be able to consolidate its own linting rules with reports from third party tools (PyLint was specifically mentioned, which is what NAV uses/used to use), and I would very much like to get this to work for all our projects at some point.

_Originally posted by @lunkwill42 in https://github.com/Uninett/zino/issues/3#issuecomment-1534202028_

johannaengland:
Relevant link: 
https://community.sonarsource.com/t/support-ruff-linter-report-files/79714/11

lunkwill42:
We should maybe consider switching to https://github.com/oxsecurity/megalinter as well - it appears to be a vastly improved and popular hard-fork of the super-linter.  It appears to support ruff out of the box.

johannaengland:
I had a look at it already and it seems very useful for our case. The only thing that I was uncertain about is if we can use it since it has the AGPL-3.0 license and Zino is under Apache 2.0. I am not well versed enough in licenses to know if that matters if we just use megalinter as a tool.

lunkwill42:
The license terms of megalinter will not apply to Zino, as no Zino code will link with megalinter code.

PR #170, opened 2024-01-29
Add ruff to Github actions

johannaengland:


lunkwill42:
I assume this is a path to begin solving what #10 is all about :)

lunkwill42 (approved):
:+1:

PR #172, opened 2024-01-30
Rename BGP event attributes to follow Zino1 protocol

### LINKED ISSUE #171: BGP: Factor out event creation

johannaengland:
I used this as an opportunity to also fix #171, which made the rest of the task much easier.

While I was working on that an comparing between Zino1 and my code I realized that I forgot the part that admin up will only update existing event, not create new events. That is also added here. 

This difference between how BGP events look in Zino1 and Zino2 was so nicely pointed out to me by @runborg.

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -309,7 +309,7 @@ class PortStateEvent(Event):
 
 class BGPEvent(Event):
     type: Literal["bgp"] = "bgp"
-    remote_address: Optional[IPAddress] = None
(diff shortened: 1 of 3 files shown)


lunkwill42 (approved):
Looks fine to me, and nice factoring out of event creation.

Reminds me again that we should look in Pydantic attributes aliases, so maybe we can have more meaningful internal names in Zino 2.0, even though names in the API need to remain backwards compatible...
https://docs.pydantic.dev/latest/concepts/alias/

PR #174, opened 2024-03-11
Add state converter from zino1 to zino2

### LINKED ISSUE #66: Write a state dump converter for Zino 1 to Zino 2
Organizations like Sikt and NORDUnet have been running Zino 1 in production for years.  At some point, they will migrate to the new Zino 2.0, and will have the need to migrate existing state data from Zino 1 to Zino 2.

Zino 1 saves its state as pure Tcl code.  This code is quite simply a bunch of assignments to global dictionaries that hold all the state data, and is imported as a Tcl module as Zino starts.  Since it doesn't contain any Tcl logic outside of simple assignments, it should be relatively easy to write a Python parser for this file, that can load the information into Zino 2.0 data structures and then dump these to disk before Zino 2.0 is started.

Since this task was not part of the original project specification, we should leave it to the very end, until we either are sure we have the person-hours to spare, or otherwise get some spending approval from somewhere.

stveit:
## Scope and purpose

Fixes #66 

Parses `.tcl` state file from Zino1 and tries to generate a Zino2 equivalent state. Has alot of placeholder functions for stuff that should eventually be supported and some stuff that maybe we will never support? I added a function for every unique command in the state dump I was working with so we are at least aware of them and can decide which ones are important.

### This pull request
* adds a `zinoconv` command line program

## Contributor Checklist

  * [ ] #310


lunkwill42 (approved):
I've cleaned up a few timezone-related things, as well as added a section about `zinoconv` (which I'm now calling it) to the README.md.  Additionally, I've added #310 to follow up on test coverage.

At the moment, I am actually unable to run `zinoconv` successfully with my old `zino-state.tcl` copy in my source tree, but I do not have the time or the resources to look into the error before the next beta release, so this will have to do for now.

Issue #179, opened 2024-03-11
Split BGP monitor task into type specific classes


johannaengland:
To make the BGP state monitor task cleaner and more readable it might be helpful to split it up into three different classes (juniper, cisco and general) and have those implement the type specific functions (`get_bgp_info`, `get_local_as`, etc.).

PR #181, opened 2024-03-12
BGP: Minimize dragging keys of sparsewalk response along

johannaengland:
This is as preparation for #179 and is one step in fixing #180 (the rest will be done when working on #179).


lunkwill42 (approved):
Yeah, it's nice to not pass around low-level PySNMP too much.  One day we might want to switch SNMP libraries, and the fewer contact points we have with PySNMP internals, the better.

PR #183, opened 2024-03-13
Add one function to parse all weird IP formats

### LINKED ISSUE #177: Unify IP address handling
In the BGP state monitor and the BFD state monitor there is a lot of doubling code for handling IP addresses/bringing them into one unified format. This should be factored out. 

The rest of Zino should also be checked if there is any other occurrence where we can factor out IP address handling.

stveit:
Fixes #177

### DIFF
--- a/src/zino/utils.py
+++ b/src/zino/utils.py
@@ -1,0 +1,27 @@
+from ipaddress import ip_address
+
+from pyasn1.type.univ import OctetString
+
+from zino.statemodels import IPAddress
+
+
+def parse_ip(ip: str) -> IPAddress:
+    """Parses IPv4 and IPv6 addresses in different formats"""
+    try:
+        return ip_address(ip)
+    except ValueError:
+        if ip.startswith("0x"):
+            return _parse_hexa_string_ip(ip)
+        if ":" in ip:
+            return _parse_colon_separated_ip(ip)
+        raise ValueError(f"Input {ip} could not be converted to IP address.")
+
+
+def _parse_hexa_string_ip(ip: str) -> IPAddress:
+    """Parses IP addresses formatted as hexastrings, e.g. 0x7f000001"""
+    return ip_address(bytes(OctetString(hexValue=ip[2:])))
+
+
+def _parse_colon_separated_ip(ip: str) -> IPAddress:
+    """Parses IP addresses formatted with a colon symbol separating every octet, e.g. 7F:00:00:01"""
+    return ip_address(bytes(OctetString(hexValue=ip.replace(":", ""))))
--- a/tests/tasks/test_bfdtask.py
+++ b/tests/tasks/test_bfdtask.py
@@ -17,7 +17,6 @@ class TestJuniper:
             "up",
             bfd_state.session_discr,
             "0x7f000001",
-            "ipv4",
         )
         assert state == bfd_state
 
--- a/tests/tasks/test_bfdtask.py
+++ b/tests/tasks/test_bfdtask.py
@@ -1,4 +1,3 @@
-import ipaddress
 from ipaddress import IPv4Address
 
 import pytest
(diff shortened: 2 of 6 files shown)


src/zino/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    """Parses IPv4 and IPv6 addresses in different formats"""
+    try:
+        return ip_address(ip)
+    except ValueError:
+        if ip.startswith("0x"):
+            return _parse_hexa_string_ip(ip)
+        if ":" in ip:
+            return _parse_colon_separated_ip(ip)
+        else:
+            raise ValueError(f"Input {ip} could not be converted to IP address.")
```
johannaengland, line 18:
```suggestion
        raise ValueError(f"Input {ip} could not be converted to IP address.")
```

src/zino/utils.py
```diff
@@ -0,0 +1,39 @@
...
+        raise ValueError(f"Input {ip} could not be converted to an IP address.")
+    return ip_address(address_str)
+
+
+def _parse_colon_separated_ip(ip: str) -> IPAddress:
+    """Parses IP addresses formatted with a colon symbol separating every octet, e.g. 7F:00:00:01
+    Works for both IPv6 and IPv4 addresses with the same format
+    """
+    address_str = bytes(int(i, 16) for i in ip.split(":"))
+    return ip_address(address_str)
```
johannaengland, line 39:
```suggestion
    return ip_address(bytes(OctetString(hexValue=ip.replace(":", ""))))
```

src/zino/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    """Parses IP addresses formatted as hexastrings, e.g. 0x7f000001"""
+    if len(ip) == 10:
+        # IPv4 address
+        address_str = ".".join((map(str, OctetString(hexValue=ip[2:]).asNumbers())))
+    elif len(ip) == 34:
+        # IPv6 address
+        address_str = ":".join(["".join(item) for item in zip(*[iter(ip[2:])] * 4)])
+    else:
+        raise ValueError(f"Input {ip} could not be converted to an IP address.")
+    return ip_address(address_str)
```
johannaengland, line 31:
```suggestion
    return ip_address(bytes(OctetString(hexValue=ip[2:])))
```

johannaengland (changes requested):
I found a way to unify this even more!

johannaengland:
And if you don't change the IP address handling in the BFD/BGP monitor tasks in this PR, then please make a new issue so that this is not forgotten :smile:

stveit:
Made BFD and BGP task use `parse_ip`. Removed using the AddrType value in BDF to decide ipv4/ipv6 since this will be detected automatically anyways. This does mean it will accept Ipv4 adresses even if the AddrType is set to Ipv6 in the case there is some error in the MIB values, but that basically just makes the code more robust. 

Does anyone have a strong opinion on using AddrType to decide ipv4/ipv6 vs letting it be automatically detected?

tests/utils_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_should_parse_colon_separated_ipv4_octets_correctly(self):
+        ip = parse_ip("7f:00:00:01")
+        assert ip == IPv4Address("127.0.0.1")
+
+    def test_should_parse_colon_separated_ipv6_octets_correctly(self):
+        ip = parse_ip("13:c7:db:1c:44:30:c8:26:63:33:ae:d0:e6:05:3a:3b")
+        assert ip == IPv6Address("13c7:db1c:4430:c826:6333:aed0:e605:3a3b")
+
+    def test_should_raise_valueerror_if_invalid_ip_format(self):
```
stveit, line 35: Imma comment this first lol 
```suggestion
    def test_should_raise_error_if_invalid_ip_format(self):
```

tests/tasks/test_bfdtask.py
```diff
@@ -10,34 +9,6 @@
 from zino.tasks.bfdtask import BFDTask
 
 
-class TestConvertAddress:
-    def test_converts_bytes_to_correct_ipv6_address(self):
-        parsed_address = BFDTask._convert_address(
-            b"\x00\x00\x00\x00\x00\x00\x00\x00\00\x00\x00\x00\x00\x00\x00\x01",
```
johannaengland, line 16: This case could be added to the utils test

johannaengland (approved):
I'm happy with the functionality. Just one comment to maybe add even more tests.

tests/tasks/test_bfdtask.py
stveit: the new parse_ip func wouldnt handle bytes like that. the old _convert_address function took bytes since thats what Iconverted "0x7f000001" format into first, then from bytes to IPAddress. I dont think anything else IP related use bytes like this

tests/utils_test.py
stveit: or of you 😨

tests/tasks/test_bfdtask.py
johannaengland: Ah, goody then, we can always add to the function if we happen to need that functionality

johannaengland (approved):

tests/tasks/test_bfdtask.py
stveit: the standard ip_address should handle bytes just fine anyways :)

Issue #173, opened 2024-03-11
Implement SNMP trap reception mechanism


lunkwill42:
### Design considerations

Zino 1 uses a trap daemon model taken from Scotty, wherein an SNMP trap multiplexer daemon (`straps` or `nmtrapd`) is run as root, and any Scotty program (like Zino) that wants a copy of all received traps can run as unprivileged users and connect to a UNIX socket to receive a  stream of the received trap messages.

While this is a really nice model for both debugging and for having multiple small and independent programs that can run on the same machine and react to traps, we don't see this as an essential model to replicate.

In Zino 2, we will instead opt for a model where the Zino program itself will listen for incoming SNMP trap messages (on UDP port 162).  This does have the side-effect that the Zino program must be started by root, and needs to be able to switch to a less privileged user immediately after opening the privileged port 162.

A model where Zino connects to a trap message multiplexer can be added later if deemed necessary.

### Sub-tasks

- [x] #189 
- [x] #193
- [x] Traps that are handled by the original Zino in some way or the other

  - These traps cause Zino to take action:
    - [x] #232
    - [x] #292
    - [x] #288

  - These traps are only parsed and logged (in simple or less simple ways):
    - [x] #226
    - [x] #319

  - These traps seem to be just ignored entirely:
    - [x] `bgpBackwardTransition` (a.k.a. `BGP4-MIB::bgpBackwardTransition`)
    - [x] `BGP4-MIB::bgpBackwardTransNotification`
    - [x] `jnxBgpM2Experiment.1.1.0.2` (a.k.a `BGP4-V2-MIB-JUNIPER::jnxBgpM2BackwardTransition`)
    - [x] `SNMPv2-MIB::authenticationFailure`
    - [x] `CISCOTRAP-MIB::tcpConnectionClose` (but contains some commented out logging code)
    - [x] `BGP4-V2-MIB-JUNIPER::jnxBgpM2BackwardTransition`

  - The default trap handler just logs incoming trap details as an "unknown trap"

lunkwill42:
The `straps` SNMP trap multiplexer is described here: https://linuxcertif.com/man/8/straps/  - but I am hard pressed to find the actual package or source code, at the moment...

lunkwill42:
[...]

PR #189, opened 2024-03-18
Add initial trap receiver functionality to Zino

lunkwill42:
This adds a simple framework for receiving and logging incoming traps, nothing more.

It also adds command line options to enable selection of trap port (optionally switching off traps entirely), and to drop privileges to a specific user if running as root.

### DIFF
--- a/src/zino/trapd.py
+++ b/src/zino/trapd.py
@@ -1,0 +1,48 @@
+"""SNMP trap 'daemon' for Zino 2"""
+import asyncio
+import logging
+
+from pysnmp.carrier.asyncio.dgram import udp
+from pysnmp.entity import config, engine
+from pysnmp.entity.rfc3413 import ntfrcv
+
+_logger = logging.getLogger(__name__)
+
+
+class TrapReceiver:
+    """Zino Adapter for SNMP trap reception using PySNMP"""
+
+    def __init__(self, address: str = "0.0.0.0", port: int = 162, loop=None):
+        self.transport = None
+        self.address = address
+        self.port = port
+        self.loop = loop if loop else asyncio.get_event_loop()
+        self.snmp_engine = engine.SnmpEngine()
+        self._communities = set()
+
+    async def open(self):
+        self.transport = udp.UdpTransport(loop=self.loop).openServerMode((self.address, self.port))
+        # This attribute needs to be awaited to ensure the socket is really opened,
+        # unsure of why PySNMP doesn't do this, or how else it's supposed to be achieved
+        await self.transport._lport
+        _logger.info("Listening for incoming SNMP traps on %r", (self.address, self.port))
+        config.addTransport(self.snmp_engine, udp.domainName + (1,), self.transport)
+
+        ntfrcv.NotificationReceiver(self.snmp_engine, self.trap_received)
+        self.snmp_engine.transportDispatcher.jobStarted(1)  # this job would never finish
+
+    def add_community(self, community: str):
+        """Adds a new community string that will be accepted on incoming packets"""
+        if community in self._communities:
+            return
+        self._communities.add(community)
+        config.addV1System(self.snmp_engine, str(len(self._communities)), community)
+
+    def trap_received(self, snmp_engine, state_reference, context_engine_id, context_name, var_binds, callback_context):
+        _logger.info(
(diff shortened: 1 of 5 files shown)


hmpf:
How about toggling traps in config? It is nice being able to do the same things both in config and cli.

src/zino/trapd.py
```diff
@@ -0,0 +1,48 @@
...
+
+    async def open(self):
+        self.transport = udp.UdpTransport(loop=self.loop).openServerMode((self.address, self.port))
+        # This attribute needs to be awaited to ensure the socket is really opened,
+        # unsure of why PySNMP doesn't do this, or how else it's supposed to be achieved
+        await self.transport._lport
+        _logger.info("Listening for incoming SNMP traps on %r", (self.address, self.port))
+        config.addTransport(self.snmp_engine, udp.domainName + (1,), self.transport)
+
+        ntfrcv.NotificationReceiver(self.snmp_engine, self.trap_received)
```
hmpf, line 31: There are no other receivers in the method, why not just "receiver"?

src/zino/trapd.py
```diff
@@ -0,0 +1,48 @@
...
+        self.snmp_engine.transportDispatcher.jobStarted(1)  # this job would never finish
+
+    def add_community(self, community: str):
```
hmpf, line 34: Will Zino 2 support SNMP 3? Might be easier learning how to doit *here* then use that kmowledge in Nav than the other way around.

hmpf (commented):
See mostly irrelevant questions, I don't know enough about traps to help with the networking bits. The python seems fine.

src/zino/trapd.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, address: str = "0.0.0.0", port: int = 162, loop=None):
+        self.transport = None
+        self.address = address
+        self.port = port
+        self.loop = loop if loop else asyncio.get_event_loop()
+        self.snmp_engine = engine.SnmpEngine()
+        self._communities = set()
+
+    async def open(self):
+        self.transport = udp.UdpTransport(loop=self.loop).openServerMode((self.address, self.port))
```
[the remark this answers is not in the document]
lunkwill42: Are you commenting on line 24 or on 27 (which, IMNSHO, is the evil part here)?

src/zino/trapd.py
lunkwill42: Huh?

src/zino/trapd.py
lunkwill42: We have no (current) plans to support SNMPv3, just to have the same feature set as Zino 1.

Multiple "systems" can be configured on the SNMP engine.  This code just adds a v1 system, which is the basic "community" for authentication, which works for both v1 and v2.  PySNMP also has the option to add v3 systems to the engine, which are more involved.

However, none of this is very relevant to NAV, since NAV uses an entirely different SNMP library.

lunkwill42:
> How about toggling traps in config? It is nice being able to do the same things both in config and cli.

Yup, I agree it would be nice, but we haven't actually implemented a generic Zino config file yet, just the `polldevs.cf` config file.  We may need a separate PR for this, as the original zino does actually have a few items in a generic config file.

src/zino/trapd.py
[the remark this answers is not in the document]
hmpf: I'm complaining about the udp-class from pysnmp, notably the naming convention of methods. Currently, the line I'm complaining about is no. 31 (it was no. 24). Who knows what number the line will have tomorrow? *shrug*

src/zino/trapd.py
hmpf: Are we not being shown the same lines again?

Never mind, it's pysnmp to blame again.

src/zino/trapd.py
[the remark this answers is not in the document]
lunkwill42: It still says you commented on line 24 here.  Not sure why your eyes need washing.  A couple of lines down is the line where eyes need washing and large amounts of morphine is required..

hmpf (approved):

lunkwill42:
This is the initial part of fixing #173, so I'm adding this comment to link these issues.

PR #192, opened 2024-03-20
Do not include event ids of closed events in CASEIDS command

### LINKED ISSUE #187: Not report cases in closed state when a new clients connects
This is a inconsistency between zino1 and zino2.
When a new client connects to the zino2 server, do not expose the client for cases in closed state.

The list reported by the `caseids` should only include active cases, all cases in closed state should be "hidden" when using this command.

johannaengland:
Closes #187.

Needs to be first checked if this is the same behavior as with Zino 1.0.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -3,6 +3,7 @@
 The Legacy API from the Tcl-based Zino 1.0 is a 'vaguely SMTP-esque line-based text protocol'.  This module
 implements this protocol using asyncio semantics.
 """
+
 import asyncio
 import inspect
 import logging
@@ -233,8 +234,10 @@ class Zino1ServerProtocol(Zino1BaseServerProtocol):
     @requires_authentication
     async def do_caseids(self):
         self._respond(304, "list of active cases follows, terminated with '.'")
-        for event_id in sorted(self._state.events.events):
-            self._respond_raw(str(event_id))
+        events = self._state.events.events
+        for event_id, event in sorted(events.items()):
+            if event.state != EventState.CLOSED:
+                self._respond_raw(str(event_id))
         self._respond_raw(".")
 
     def _translate_case_id_to_event(responder: callable):  # noqa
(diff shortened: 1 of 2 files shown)


lunkwill42:
> Needs to be first checked if this is the same behavior as with Zino 1.0.

Seems consistent, according to my tests.  The main difference is that Zino 1 doesn't filter the events in the implementation of `CASEIDS`, rather it only indexes non-closed events.

However, it doesn't really matter how this works internally in Zino 2 as long as `CASEIDS` doesn't list closed events, and closed events cannot be reopened.

src/zino/api/legacy.py
```diff
@@ -233,8 +234,11 @@ async def do_version(self):
     @requires_authentication
     async def do_caseids(self):
         self._respond(304, "list of active cases follows, terminated with '.'")
-        for event_id in sorted(self._state.events.events):
-            self._respond_raw(str(event_id))
+        events = self._state.events.events
+        event_ids = sorted(events)
+        for event_id in event_ids:
+            if events[event_id].state is not EventState.CLOSED:
+                self._respond_raw(str(event_id))
```
lunkwill42, line 241:
```suggestion
        events = self._state.events.events
        for event_id, event in sorted(events.items()):
            if event.state != EventState.CLOSED:
                self._respond_raw(str(event_id))
```

lunkwill42: Alternatively, the `Events` class could gain methods that do the filtering on closed/non-closed states - but let's do that only if we see that we need this in more places.

lunkwill42, line 241:
```suggestion
        events = self._state.events.events
        open_event_ids = (event.id for event in events.values() if event.state != EventState.CLOSED)
        for event_id in sorted(open_event_ids):
            self._respond_raw(str(event_id))
```

lunkwill42 (approved):
Approved, but with two alternate suggestion for a slightly cleaner version of the filter loop.

I have confirmed that this should correspond to Zino 1 behavior.

PR #193, opened 2024-03-22
 Implement a generic trap dispatcher framework in `zino.trapd.TrapReceiver`

lunkwill42:
This PR aims to create a generic framework for dispatching incoming traps in the `TrapReceiver` class.  Features that are planned:

- [x] Do the "generic" work of preprocessing a trap: 
  - [x] Map it to a known poll device, otherwise drop it.  
  - [x] Translate names and data on incoming varbinds.
  - [x] Drop the trap if it is missing required fields (like trap-OID or sysUpTime, see `trap.tcl` in Zino 1 for details)
- [x] Provide a method to register functions that are interested in handling specific traps
- [x] Dispatch traps to registered trap handlers
- [x] Log the contents of traps that weren't dispatched to any handlers.
- [x] Ignore same set of traps as ignored by Zino 1

### DIFF
--- a/src/zino/trapd.py
+++ b/src/zino/trapd.py
@@ -3,7 +3,7 @@ import asyncio
 import logging
 from dataclasses import dataclass, field
 from ipaddress import ip_address
-from typing import Any, List, NamedTuple, Optional, Protocol
+from typing import Any, List, NamedTuple, Optional, Protocol, Set
 
 from pysnmp.carrier.asyncio.dgram import udp
 from pysnmp.entity import config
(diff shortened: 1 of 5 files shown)


lunkwill42:
factored out 1a3ca48d677b28f4699db93df70807bf8bd4d716 in the latest force-push

src/zino/trapd.py
```diff
@@ -97,6 +105,8 @@ def __init__(
         self._communities = set()
         self._observers: dict[TrapType, List[TrapObserver]] = {}
 
+        self.observe(ignore_trap, *IGNORED_TRAPS)
```
johannaengland, line 108: Oh, this is smooth, I like it!

johannaengland (approved):

hmpf (approved):

src/zino/trapd.py
stveit: What is the benefit of doing this instead of adding a check to the `dispatch_traps` function? something like

```
if (trap.mib, trap.name) in IGNORED_TRAPS:
    return
```

lunkwill42: Separation of concerns, really.  If the ignore procedure needs to change somehow (like adding logging or "whatever"), only the `ignore_trap` function needs to change, not the TrapReceiver.

Ideally, the ignore-observer would be registered with the trapreceiver externally, but I couldn't figure out where it would fit in without putting a lot of stuff into `zino.zino`.

Issue #101, opened 2023-10-18
Implement the `PM` family of API commands


lunkwill42:
The `PM` API command is used to manipulated planned maintenance entries in Zino (as implemented in #194)

However, the `PM` command feature multiple subcommands, which will necessitate changes to how the `Zino1BaseServerProtocol` dispatches commands to methods.  Example from Zino 1 server session:

```
PM HELP
200- PM subcommands are:
200  ADD LIST CANCEL DETAILS MATCHING ADDLOG LOG
```

### Subtasks

- [x] #255
- [x] Implement the `PM ADD` API subcommand
- [x] Implement the `PM LIST` API subcommand
- [x] Implement the `PM CANCEL` API subcommand
- [x] Implement the `PM DETAILS` API subcommand
- [x] Implement the `PM MATCHING` API subcommand
- [x] Implement the `PM ADDLOG` API subcommand
- [x] Implement the `PM LOG` API subcommand

PR #194, opened 2024-03-24
Add planned maintenance

### LINKED ISSUE #61: Add planned maintenance (PM) calendar to Zino backend state
Planned Maintenance needs a data model in Zino 2.  A PM tells Zino to automatically put events into `ignored` state when they are created, if they match an existing/active PM.

This includes designing where the PM match should take place in the code, and also discovering what is included in the original PM  model of Zino 1.

The API bit is tracked separately: 
- [ ] #101

stveit:
Fixes #61, API still missing (tracked by #101). That will be separate PR


johannaengland (changes requested):

johannaengland:
I just checked in Zino1.0 and found that `get_started_planned_maintenances` will actually not check if the pms have ended before the current time. I think this might be something to discuss, even though it is quite an edge case, since we don't think any pms will have started after the last run, but ended before the time the function is called.

src/zino/planned_maintenance.py
```diff
@@ -110,44 +119,95 @@ def periodic(self):
...
+    def _start(self, pm: PlannedMaintenance):
         from zino.state import state
 
-        event.state = EventState.IGNORED
-        state.events.commit(event)
+        events = self._get_or_create_events(pm)
+        for event in events:
+            event.state = EventState.IGNORED
+            state.events.commit(event)
+            pm.event_ids.append(event.id)
```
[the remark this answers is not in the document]
stveit: thats my understanding of how it works, since that will catch any alarm that occurs during the PM timeframe even if it didnt exist before. tcl isnt the easiest to read tho..

src/zino/planned_maintenance.py
```diff
@@ -110,44 +119,95 @@ def periodic(self):
...
-        state.events.commit(event)
+        if event.state in [EventState.IGNORED, EventState.CLOSED]:
+            return
+
+        active_pms = self.get_active_planned_maintenances()
+        for pm in active_pms:
+            if pm.matches(event):
+                event.state = EventState.IGNORED
+                event.add_log(f"entered into existing active PM event id {pm.id}")
+                state.events.commit(event)
```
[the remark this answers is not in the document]
stveit: yap that should be there

src/zino/planned_maintenance.py
```diff
@@ -99,6 +99,42 @@ def _call_observers(self) -> None:
         for observer in self._observers:
             observer()
 
+    def periodic(self):
```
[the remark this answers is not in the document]
stveit: Yes you can call it whatever you want :) called it same as zino1 since it wasnt obvious to me what the best name would be

src/zino/planned_maintenance.py
```diff
@@ -68,21 +68,27 @@ def get_started_planned_maintenances(self, now: datetime.datetime) -> list[Plann
...
         """
-        # Note: last run (self.last_run) can be None, which would mean this task has never run before
-        pass
+        if self.last_run:
+            return [pm for pm in self.planned_maintenances if self.last_run < pm.end_time <= now]
+        else:
+            return [pm for pm in self.planned_maintenances if pm.end_time <= now]
 
-    def get_active_planned_maintenances(self, now: datetime.datetime) -> list[PlannedMaintenance]:
+    def get_active_planned_maintenances(self) -> list[PlannedMaintenance]:
```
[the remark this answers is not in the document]
stveit: Decided to do it same as zino1 does, since it can technically change the behavior. But I agree that it is silly

src/zino/planned_maintenance.py
```diff
@@ -110,44 +119,95 @@ def periodic(self):
...
-        ended_maintenances = self.get_ended_planned_maintenances(now=now)
-        for ended_pm in ended_maintenances:
-            for event in ended_pm.pm_events:
-                if event.state is not EventState.OPEN:
-                    self._unignore_event(event)
-
-    def _ignore_event(self, event: Event):
+            self._check_event(event)
+
+        # End a PM and set events matching the PM to open
```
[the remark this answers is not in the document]
stveit: I guess, could change to something like "set events matching the ended PM to open"

stveit:
> I just checked in Zino1.0 and found that `get_started_planned_maintenances` will actually not check if the pms have ended before the current time. I think this might be something to discuss, even though it is quite an edge case, since we don't think any pms will have started after the last run, but ended before the time the function is called.

yeah noticed this too, seems like an oversight but I guess no planned maintenance has ever been shorter than whatever interval the PM stuff runs at

johannaengland:
> > I just checked in Zino1.0 and found that `get_started_planned_maintenances` will actually not check if the pms have ended before the current time. I think this might be something to discuss, even though it is quite an edge case, since we don't think any pms will have started after the last run, but ended before the time the function is called.
> 
> yeah noticed this too, seems like an oversight but I guess no planned maintenance has ever been shorter than whatever interval the PM stuff runs at

Yes, that would make very little sense to have a PM shorter than one minute. 
But I still think we can add that check, that makes testing easier and seems like good practice to me.

src/zino/planned_maintenance.py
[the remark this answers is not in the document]
johannaengland: I would change it then, this potential change in behavior can invite bugs, so let's avoid that

src/zino/planned_maintenance.py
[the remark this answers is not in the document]
johannaengland: No, I think this is how it is done today. So we should do it like this as well, but put the eventual improvement on the list for further development

src/zino/planned_maintenance.py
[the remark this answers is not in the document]
stveit: I think its the opposite, not using the same date everywhere could potentially invite bugs :P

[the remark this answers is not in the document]
johannaengland: I might've expressed myself not clearly enough, I also meant that we should use the same date everywhere :D

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+        This means it has started before `now` and the end_time is later than `now`
+        """
+        return [pm for pm in self.planned_maintenances.values() if pm.start_time < now < pm.end_time]
+
+    def get_old_planned_maintenances(self, now: datetime.datetime) -> list[PlannedMaintenance]:
+        """Returns all planned maintenances that should get deleted
+
+        This means that `now` is 3 days later than end_time
+        """
+        return [pm for pm in self.planned_maintenances.values() if now - pm.end_time > datetime.timedelta(days=3)]
```
lunkwill42, line 112: Magic number alert.  Not sure if the number is configurable in Zino 1, but the timedelta should be defined as a constant somewhere, e.g.

```python
PM_EXPIRY_TIME = datetime.timedelta(days=3)
```

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+    def get_next_available_pm_id(self) -> int:
+        """Returns the next available planned maintenance id"""
+        self.last_pm_id += 1
+        return self.last_pm_id
+
+    def _call_observers(self) -> None:
+        for observer in self._observers:
+            observer()
+
+    def periodic(self, state: "ZinoState"):
```
lunkwill42, line 129: I agree this name should be more descriptive, but naming things is hard :-P

Random suggestions could be  `update_pm_calendar` or `update_pm_states`

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+        now = datetime.datetime.now()
+
+        # Initiate PM once it becomes active
+        for started_pm in self.get_started_planned_maintenances(now=now):
+            self._start(state, started_pm)
+
+        # Make sure all events that matches a PM is ignored
```
lunkwill42, line 136:
```suggestion
        # Make sure all events that match a PM is ignored
```

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+        events = self._get_or_create_events(state, pm)
+        for event in events:
+            # get special handling of embryonic -> open transition first
+            state.events.commit(event)
+
+            event.state = EventState.IGNORED
+            state.events.commit(event)
+            pm.event_ids.append(event.id)
+
+    def _check_event(self, state: "ZinoState", event: Event, now: datetime.datetime):
```
lunkwill42, line 160: What does `_check_event` really mean?

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+        for ended_pm in self.get_ended_planned_maintenances(now=now):
+            self._end(state, ended_pm)
+
+        old_pms = self.get_old_planned_maintenances(now)
+        for pm in old_pms:
+            self.close_planned_maintenance(pm.id, "timer expiry for old PMs", "zino")
+
+        self.last_run = now
+
+    def _start(self, state: "ZinoState", pm: PlannedMaintenance):
```
lunkwill42, line 150: Start what?

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+
+        active_pms = self.get_active_planned_maintenances(now)
+        for pm in active_pms:
+            if pm.matches_event(event, state):
+                event.state = EventState.IGNORED
+                event.add_log(f"entered into existing active PM event id {pm.id}")
+                state.events.commit(event)
+                pm.event_ids.append(event.id)
+
+    def _end(self, state: "ZinoState", pm: PlannedMaintenance):
```
lunkwill42, line 172: End what?

lunkwill42 (changes requested):
I don't know the original PM code intimately, so I'm just going by intuition on some of the functionality.

I'm left with a feeling that perhaps more of the functionality of the `PlannedMaintenances` class could be moved to the `PlannedMaintenance` class, especially much of the things that just take a single `PlannedMaintenance` object and doesn't really touch the internal state of the `PlannedMaintenances` object itself.

Another idea could be to split `PlannedMaintance` class could even be split into subclasses, one for each type ("portstate", "device").  The event matching or generating methods could then be subclass specific.

Maybe an IRL pow-wow would be better here :)

src/zino/statemodels.py
```diff
@@ -346,3 +350,79 @@ class AlarmEvent(Event):
...
+        Returns true if `string` matches `pattern`.
+        """
+        pass
+
+    def _regex_match(self, pattern: str, string: str) -> bool:
+        """Matches `string` against regex expression `pattern`.
+        Returns true if there is a match.
+        """
+        compiled_pattern = re.compile(pattern)
+        return bool(compiled_pattern.match(string))
```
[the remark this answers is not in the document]
stveit: can be static, but since theyre only used internally in this model I would prefer keeping them inside the class at least

src/zino/planned_maintenance.py
stveit: this is moved into Model class. The name remains but makes more sense in the context of being used on an PM object

src/zino/planned_maintenance.py
stveit: this is moved into Model class. The name remains but makes more sense in the context of being used on an PM object

src/zino/planned_maintenance.py
stveit: renamed and refactored a bit

src/zino/planned_maintenance.py
stveit: renamed

src/zino/statemodels.py
```diff
@@ -346,3 +350,79 @@ class AlarmEvent(Event):
...
     def subindex(self) -> SubIndex:
         return self.alarm_type
+
+
+class PlannedMaintenance(BaseModel):
+    id: Optional[int] = None
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    type: Literal["portstate", "device"]
+    match_type: Literal["regexp", "str", "exact", "intf-regexp"]
```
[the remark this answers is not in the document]
stveit: added enums

stveit:
#234 #235

stveit:
#236

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,217 @@
...
+    def close_planned_maintenance(self, id: int, reason: str, user: str) -> None:
+        """Deletes planned maintenance with the given id"""
+        # See `close` function in Zino 1.0 `pm.tcl`
+        pm = self.planned_maintenances.get(id, None)
+        if not pm:
+            # TODO figure out if this is enough
+            return
+        pm.add_log(f"PM closed by {user}: {reason}")
+        del self.planned_maintenances[id]
+        self._call_observers()
```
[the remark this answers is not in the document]
lunkwill42: Conclusion here is that this is *unimplemented* in Zino 1 (feature only exists in the comments), so we added #235

src/zino/statemodels.py
```diff
@@ -363,3 +368,156 @@ class AlarmEvent(Event):
...
+        """
+        return fnmatch.fnmatch(string, pattern)
+
+    @staticmethod
+    def _regex_match(pattern: str, string: str) -> bool:
+        """Matches `string` against regex expression `pattern`.
+        Returns true if there is a match.
+        """
+        compiled_pattern = re.compile(pattern)
+        return bool(compiled_pattern.match(string))
```
lunkwill42, line 441: I would move these to the bottom of the file.  I'm a proponent of the step-down rule.  A code file should read as higher-order concepts first, nitty-gritty details at the bottom (where the constraints of the programming language doesn't prevent you from using that order, that is)

lunkwill42, line 441: Compiling the pattern before using it here makes no sense.  The compiling step makes sense if we want to optimize performance by reusing the compiled pattern over and over, but here it's essentially throwing away the compiled pattern after each use.

Either use `re.match()` directly, or ensure that the PM classes pre-compile any regex-patterns so they can be re-used.

src/zino/statemodels.py
```diff
@@ -363,3 +368,156 @@ class AlarmEvent(Event):
...
+
+
+class DeviceMaintenance(PlannedMaintenance):
+    type: PmType = PmType.DEVICE
+
+    def matches_event(self, event: Event, state: "ZinoState") -> bool:
+        """Returns true if `event` will be affected by this planned maintenance"""
+        if event.type not in ["reachability", "alarm"]:
```
lunkwill42, line 449:
```suggestion
        if not isinstance(event, (ReachabilityEvent, AlarmEvent)):
```

src/zino/statemodels.py
```diff
@@ -363,3 +368,156 @@ class AlarmEvent(Event):
...
+            return self.match_expression == device.name
+        return False
+
+    def _get_or_create_events(self, state: "ZinoState") -> list[Event]:
+        """Creates/gets events that are affected by the given starting planned
+        maintenance
+        """
+        events = []
+        # all devices that the pm should affect
+        devices = [device for device in state.devices.devices.values() if self.matches_device(device)]
```
lunkwill42, line 472: This doesn't need to keep a full list in memory, so it could simply use a generator expression here.
```suggestion
        devices = (device for device in state.devices.devices.values() if self.matches_device(device))
```

src/zino/statemodels.py
```diff
@@ -363,3 +368,156 @@ class AlarmEvent(Event):
...
+            events.append(event)
+        return events
+
+    def _get_matching_ports(self, state: "ZinoState") -> list[tuple[DeviceState, Port]]:
+        ports = []
+        for device in state.devices.devices.values():
+            for port in device.ports.values():
+                if self.matches_portstate(device, port):
+                    ports.append((device, port))
+        return ports
```
lunkwill42, line 523: It doesn't seem to need to keep an actual list of ports in memory here (also, multiple list appends are usually a lot slower than list comprehensions or generators). The `_get_matching_ports` function could be changed into a generator by simply `yield`ing each matched port

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+            pm_class=DeviceMaintenance,
+            match_type="str",
+            match_expression="hello",
+            match_device="device",
+        )
+        started_pms = pms.get_started_planned_maintenances(now=datetime.now())
+        assert recent_pm in started_pms
+        assert active_pm not in started_pms
+        assert ended_pm not in started_pms
+        assert old_pm not in started_pms
```
lunkwill42, line 42: This seems to pack a lot of different asserts into a single test.  It would be better if each concept was tested individually using tests with more specific names.  E.g. if a test called `test_started_planned_maintenances_should_not_contain_already_active_pms` fails, I know more about what's going on.

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+        active_pms = pms.get_active_planned_maintenances(datetime.now())
+        assert active_pm in active_pms
+        assert ended_pm not in active_pms
+        assert old_pm not in active_pms
+
+    def test_get_old_planned_maintenances(self, pms, active_pm, ended_pm, old_pm):
+        old_pms = pms.get_old_planned_maintenances(now=datetime.now())
+        assert old_pm in old_pms
+        assert active_pm not in old_pms
+        assert ended_pm not in old_pms
```
lunkwill42, line 61: Same thing here: It's best to keep the number of asserts per test limited - at least not assert multiple different behaviors in a single test.

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+class TestClosePlannedMaintenance:
+    def test_existing_pm_should_be_deleted(self, pms, old_pm):
+        pms.close_planned_maintenance(old_pm.id, "test", "test")
+        assert old_pm.id not in pms.planned_maintenances
+
+    def test_should_not_raise_exception_if_no_matching_pm(self, pms, old_pm):
```
lunkwill42, line 69:
```suggestion
    def test_when_there_are_no_matching_pms_it_should_not_raise_exception(self, pms, old_pm):
```

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def test_should_call_observers_after_closing_pm(self, pms, old_pm):
+        observer = Mock()
+        pms.add_pm_observer(observer.observe)
+        pms.close_planned_maintenance(old_pm.id, "test", "test")
+        assert observer.observe.called
+
+
+class TestUpdatePmStates:
+    def test_events_matching_active_device_pm_are_set_to_ignored(self, state, active_pm):
```
lunkwill42, line 80:
```suggestion
    def test_events_matching_active_device_pm_should_be_set_to_ignored(self, state, active_pm):
```

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+        device = state.devices.get("device")
+        state.planned_maintenances.update_pm_states(state)
+        reachability_event = state.events.get(device.name, None, ReachabilityEvent)
+        assert reachability_event.state == EventState.IGNORED
+        yellow_alarm_event = state.events.get(device.name, "yellow", AlarmEvent)
+        assert yellow_alarm_event.state == EventState.IGNORED
+        red_alarm_event = state.events.get(device.name, "red", AlarmEvent)
+        assert red_alarm_event.state == EventState.IGNORED
+
+    def test_events_matching_active_portstate_pm_are_set_to_ignored(self, state, active_portstate_pm):
```
lunkwill42, line 90:
```suggestion
    def test_events_matching_active_portstate_pm_should_be_set_to_ignored(self, state, active_portstate_pm):
```

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+        device = state.devices.get("device")
+        port = Port(ifindex=1, ifdescr="port")
+        device.ports[port.ifindex] = port
+        state.planned_maintenances.update_pm_states(state)
+        event = state.events.get(device.name, port.ifindex, PortStateEvent)
+        assert event.state == EventState.IGNORED
+
+    def test_events_affected_by_pm_get_opened_after_pm_ends(self, state, ended_pm):
```
lunkwill42, line 98:
```suggestion
    def test_when_pm_ends_its_affected_events_should_be_opened(self, state, ended_pm):
```

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+        # Create event and register it as affected by PM
+        reachability_event = state.events.create_event(device.name, None, ReachabilityEvent)
+        reachability_event.state == EventState.IGNORED
+        state.events.commit(reachability_event)
+        ended_pm.event_ids.append(reachability_event.id)
+
+        state.planned_maintenances.update_pm_states(state)
+        assert reachability_event.state == EventState.OPEN
+
+    def test_old_pms_are_deleted(self, state, old_pm):
```
lunkwill42, line 110:
```suggestion
    def test_old_pms_should_be_deleted(self, state, old_pm):
```

tests/planned_maintenance_test.py
```diff
@@ -0,0 +1,183 @@
...
+        assert old_pm.id in state.planned_maintenances.planned_maintenances
+        state.planned_maintenances.update_pm_states(state)
+        assert old_pm.id not in state.planned_maintenances.planned_maintenances
+
+    def test_event_opened_after_pm_was_initiated_is_set_to_ignored(self, state, active_pm):
```
lunkwill42, line 115:
```suggestion
    def test_event_opened_after_pm_was_initiated_should_be_set_to_ignored(self, state, active_pm):
```

lunkwill42 (changes requested):
This is really turning out nice!

Cannot finish without some nit-picks, though! :-)

src/zino/statemodels.py
stveit: fixed

stveit: fixed

src/zino/statemodels.py
stveit: fixed

src/zino/statemodels.py
stveit: fixed

tests/planned_maintenance_test.py
stveit: fixed

tests/planned_maintenance_test.py
stveit: fixed

src/zino/statemodels.py
stveit: fixed

src/zino/planned_maintenance.py
johannaengland: It is not configurable in Zino 1, it is directly defined in the code (see `old_pms` in `pm.tcl`)

lunkwill42 (approved):
Excellent! Let's get going with the API part! :-)

tests/zino_test.py
```diff
@@ -106,6 +106,26 @@ def test_when_less_than_10_seconds_remains_until_next_dump_it_should_not_resched
             assert not mock_job.modify.called
 
 
+class TestZinoRescheduleDumpStateOnPmChange:
+    def test_when_more_than_10_seconds_remains_until_next_dump_it_should_reschedule(self):
```
johannaengland, line 110:
```suggestion
    def test_when_more_than_10_seconds_remain_until_next_dump_it_should_reschedule(self):
```

tests/zino_test.py
```diff
@@ -106,6 +106,26 @@ def test_when_less_than_10_seconds_remains_until_next_dump_it_should_not_resched
...
+        scheduler = get_scheduler()
+        mock_job = Mock(next_run_time=now() + timedelta(minutes=5))
+
+        with patch.object(scheduler, "get_job") as get_job:
+            get_job.return_value = mock_job
+            zino.reschedule_dump_state_on_pm_change()
+            assert mock_job.modify.called
+
+    def test_when_less_than_10_seconds_remains_until_next_dump_it_should_not_reschedule(self):
```
johannaengland, line 119:
```suggestion
    def test_when_less_than_10_seconds_remain_until_next_dump_it_should_not_reschedule(self):
```

johannaengland (changes requested):

tests/statemodels/device_maintenance_test.py
```diff
@@ -0,0 +1,112 @@
...
+        start_time=datetime.datetime.now() - datetime.timedelta(days=1),
+        end_time=datetime.datetime.now() + datetime.timedelta(days=1),
+        match_type=request.param,
+        match_expression=device.name,
+        match_device=None,
+    )
+
+
+@pytest.fixture
+def matching_device_pm(device, port) -> DeviceMaintenance:
```
johannaengland, line 90: Sonarlint complains here for the type hint: `Remove this yield statement or annotate function "matching_device_pm" with "typing.Generator" or one of its supertypes.`

tests/statemodels/device_maintenance_test.py
```diff
@@ -0,0 +1,112 @@
...
+            start_time=datetime.datetime.now() - datetime.timedelta(days=1),
+            end_time=datetime.datetime.now() + datetime.timedelta(days=1),
+            match_type="str",
+            match_expression=port.ifdescr,
+            match_device=None,
+        )
+
+
+@pytest.fixture
+def nonmatching_device_pm(device, port) -> DeviceMaintenance:
```
johannaengland, line 103: Sonarlint complains here for the type hint: `Remove this yield statement or annotate function "matching_device_pm" with "typing.Generator" or one of its supertypes.`

tests/statemodels/portstate_maintenance_test.py
```diff
@@ -0,0 +1,129 @@
...
+        start_time=datetime.datetime.now() - datetime.timedelta(days=1),
+        end_time=datetime.datetime.now() + datetime.timedelta(days=1),
+        match_type=request.param,
+        match_expression=port.ifdescr,
+        match_device=device.name,
+    )
+
+
+@pytest.fixture
+def matching_portstate_pm(device, port) -> PortStateMaintenance:
```
johannaengland, line 107: Sonarlint complains here for the type hint: `Remove this yield statement or annotate function "matching_device_pm" with "typing.Generator" or one of its supertypes.`

tests/statemodels/portstate_maintenance_test.py
```diff
@@ -0,0 +1,129 @@
...
+            start_time=datetime.datetime.now() - datetime.timedelta(days=1),
+            end_time=datetime.datetime.now() + datetime.timedelta(days=1),
+            match_type="str",
+            match_expression=port.ifdescr,
+            match_device=device.name,
+        )
+
+
+@pytest.fixture
+def nonmatching_portstate_pm(device, port) -> PortStateMaintenance:
```
johannaengland, line 120: Sonarlint complains here for the type hint: `Remove this yield statement or annotate function "matching_device_pm" with "typing.Generator" or one of its supertypes.`

johannaengland (changes requested):

tests/statemodels/device_maintenance_test.py
lunkwill42: What's sonarlint?  I agree that these annotations are incorrect, but where are are you seeing this complaints, @johannaengland ?

johannaengland: In VSCode, since I have the [SonarLint extension](https://github.com/SonarSource/sonarlint-vscode)

lunkwill42: Question then become: Why doesn't our SonarCloud integration trigger the same linter warnings in the GitHub workflow? :-P

johannaengland: Very good question, I do not have access to our SonarCloud configuration to try to figure that out

lunkwill42: I can see if I can find some knobs.  Nevertheless, very good and thorough review, @johannaengland - I'll leave the rest up to @stveit once he returns :)

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,143 @@
...
+    def get_next_available_pm_id(self) -> int:
+        """Returns the next available planned maintenance id"""
+        self.last_pm_id += 1
+        return self.last_pm_id
+
+    def _call_observers(self) -> None:
+        for observer in self._observers:
+            observer()
+
+    def update_pm_states(self, state: "ZinoState"):
```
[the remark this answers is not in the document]
stveit: fixed

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,143 @@
...
+            match_expression=match_expression,
+            match_device=match_device,
+        )
+        self.planned_maintenances[pm_id] = pm
+        self._call_observers()
+        return pm
+
+    def close_planned_maintenance(self, id: int, reason: str, user: str) -> None:
+        """Deletes planned maintenance with the given id"""
+        # See `close` function in Zino 1.0 `pm.tcl`
```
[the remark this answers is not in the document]
stveit: fixed

src/zino/planned_maintenance.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Delete PMs that ended a certain amount of time ago
+        old_pms = self.get_old_planned_maintenances(now)
+        for pm in old_pms:
+            self.close_planned_maintenance(pm.id, "timer expiry for old PMs", "zino")
+
+        self.last_run = now
+
+    def _ignore_event_if_it_has_active_planned_maintenance(self, state: "ZinoState", event: Event, now: datetime):
+        active_pms = self.get_active_planned_maintenances(now)
```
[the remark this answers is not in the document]
stveit: fixed

src/zino/statemodels.py
```diff
@@ -363,3 +369,152 @@ class AlarmEvent(Event):
     @property
     def subindex(self) -> SubIndex:
         return self.alarm_type
+
+
+class MatchType(StrEnum):
+    """The set of allowable BGP oper states"""
```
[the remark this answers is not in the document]
stveit: yes

src/zino/statemodels.py
```diff
@@ -363,3 +369,152 @@ class AlarmEvent(Event):
...
+
+    REGEXP = "regexp"
+    STR = "str"
+    EXACT = "exact"
+    INTF_REGEXP = "intf-regexp"
+
+
+class PmType(StrEnum):
```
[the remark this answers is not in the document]
stveit: fixed

src/zino/statemodels.py
```diff
@@ -372,7 +372,7 @@ def subindex(self) -> SubIndex:
 
 
 class MatchType(StrEnum):
-    """The set of allowable BGP oper states"""
+    """The set of allowable match types for PlannedMaintenanc objects"""
```
johannaengland, line 375:
```suggestion
    """The set of allowable match types for PlannedMaintenance objects"""
```

johannaengland (approved):
Only one typo left, otherwise good!

PR #195, opened 2024-03-25
Fix workflow syntax and only publish test results on Uninett repo

johannaengland:
My editor Github workflow extension complained about the quotes, which is why I removed them.

And I always get an error for the "Publish test results" workflow, when pushing to the master branch of my fork, which is why I'm restricting this part to workflows on the Uninett repo.

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -10,7 +10,8 @@ jobs:
   publish-test-results:
     name: "Publish test results"
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


lunkwill42 (approved):
I believe this should work, but we should also make note of this thread in case we need it again: https://github.com/orgs/community/discussions/26829

We've had similar issues in our other repos, where the case is really that we want to differentiate between pull requests and other events, or between PRs coming from a forked repo vs one coming from the same repo.

PR #196, opened 2024-03-25
Switch from super linter to mega linter

johannaengland:
As mentioned in #10. Reverts #170.

More info to the configuration of MegaLinter can be found here: https://megalinter.io/latest/configuration/
For example that `DEFAULT_BRANCH` has been deprecated.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -2,12 +2,7 @@
 # MegaLinter GitHub Action configuration file
 # More info at https://megalinter.io
 name: MegaLinter
-
-on:
-  # Trigger mega-linter at every push to master. Action will also be visible from all Pull Requests
-  push:
-    branches: master
-  pull_request:
+on: [push, pull_request]
 
 jobs:
   megalinter:
@@ -23,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          fetch-depth: 0 # If you use VALIDATE_ALL_CODEBASE = true, you can remove this line to improve performances
+          fetch-depth: 0
       # MegaLinter
       - name: MegaLinter
         id: ml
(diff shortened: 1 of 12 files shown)


.github/workflows/linter.yml
```diff
@@ -1,74 +1,55 @@
...
+        id: ml
+        # You can override MegaLinter flavor used to have faster performances
+        # More info at https://megalinter.io/flavors/
+        uses: oxsecurity/megalinter/flavors/python@v7
         env:
-          VALIDATE_ALL_CODEBASE: false
-          DEFAULT_BRANCH: master
+          # All available variables are described in documentation
+          # https://megalinter.io/configuration/
+          VALIDATE_ALL_CODEBASE: true
```
johannaengland, line 36: This needs to be set to `false` before merging, but it is needed for testing to show that it fails before reformatting things with black

lunkwill42:
> The "Reformat with black" commit needs to be added to the `git ignore revs` file after merging this PR.

IMHO, its better to add this commit as part if *this PR*

pyproject.toml
```diff
@@ -90,6 +90,7 @@ exclude = '''
     | buck-out
     | build
     | dist
+    | src/zino/mibdumps
```
lunkwill42, line 93: :1st_place_medal:

.github/workflows/linter.yml
lunkwill42: Can't it be  dynamically set to `false` for PRs and to `true` for branch pushes?

lunkwill42 (commented):
Looks ok to me, but see inline comments.  Would also be nice if the `.git-blame-ignore-revs` was part of the PR that introduces the need for it..

.github/workflows/linter.yml
```diff
@@ -24,6 +24,13 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0 # If you use VALIDATE_ALL_CODEBASE = true, you can remove this line to improve performances
+      - name: Set environment for branch
+        run: |
+          if [[ $GITHUB_REF_NAME == 'master' ]]; then
+              echo "VALIDATE_ALL_CODEBASE: true" >> "$GITHUB_ENV"
+          else
+              echo "VALIDATE_ALL_CODEBASE: false" >> "$GITHUB_ENV"
+          fi
```
lunkwill42, line 33: We might want to validate the full codebase on any feature branch also, not just master (not that we have many feature branches yet).  However, specifically on events that were triggered from PRs, we only want the diff...

johannaengland:
I am uncertain currently if it will be possible for MegaLinter to comments a summary to a PR if that PR is coming from a fork. I think we will have to see what happens when this PR is merged.

lunkwill42 (approved):

PR #200, opened 2024-03-28
Add lastevent attribute to BFD events

### LINKED ISSUE #198: BFD: missing parameter `lastevent`
It looks like all BFD cases are missing the attribute `lastevent` . 
This makes the curitz client crash as this field is used to calculate the description field of the case.

examples of data: 
```
lastevent       : 'changed from down to up (poll)'
lastevent       : 'changed from up to down (trap) diag neighborSignaledSessionDown'
lastevent       : 'changed from up to down (trap) diag neighborSignaledSessionDown'
```

As for now it looks like to be the last part of the newest message in the system log for this case

johannaengland:
Closes #198.


stveit (approved):
looks fine

lunkwill42:
@johannaengland @stveit  I would just like to note that the original Zino 1 does not seem to put the exact same message in `lastevent` as it actually puts in the event log.

Example from the BGP code:

```tcl
    setEventAttr $eid "lastevent" "peer was reset (now up)"

    set s [format "%s peer %s AS %s was reset (now up)" \
	       $r $remote $remote_as $a_state]
    log $s
    eventLog $eid $s
```

johannaengland:
> @johannaengland @stveit I would just like to note that the original Zino 1 does not seem to put the exact same message in `lastevent` as it actually puts in the event log.
> 
> Example from the BGP code:
> 
> ```tcl
>     setEventAttr $eid "lastevent" "peer was reset (now up)"
> 
>     set s [format "%s peer %s AS %s was reset (now up)" \
> 	       $r $remote $remote_as $a_state]
>     log $s
>     eventLog $eid $s
> ```

That is true. I did exactly copy it how it is in Zino 1 for BFD and JuniperAlarm events. Should I change it in BGP as well?

lunkwill42:
> That is true. I did exactly copy it how it is in Zino 1 for BFD and JuniperAlarm events. Should I change it in BGP as well?

Please do - it's probably best to keep all output more or less the same as the original until we're asked to do it differently :)

PR #201, opened 2024-03-29
Do not allow reopening of closed events

### LINKED ISSUE #188: User should not be able to modify objects in closed state
This is a inconsistency between zino1 and zino2.
in zino1, when a case is in closed state it is complely locked for editing.

In zino2 when a ticket is in closed state the user should not be able to modify it. eg. adding history.
There could be a discussion on if it should be possible to reopen a case at all. 
There are instances where reopening a case could be beneficial, eg when closing a case by accident eg. by a mass-update.
But this again could make things like triggering an external action when closing cases more difficult as the close action could be done multiple times on a case object.

johannaengland:
Might close #188 (depending on if changing closed events should be allowed or not).

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -1,4 +1,5 @@
 """Basic data models for keeping/serializing/deserializing Zino state"""
+
 import datetime
 import logging
 from enum import Enum
@@ -15,6 +16,12 @@ AlarmType = Literal["yellow", "red"]
 SubIndex = Union[None, int, IPAddress, AlarmType]
 
 _logger = logging.getLogger(__name__)
+
+
+class ClosedEventError(Exception):
+    """Closed events cannot be reopened"""
+
+    pass
 
 
 class BGPStyle(StrEnum):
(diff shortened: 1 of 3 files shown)


src/zino/statemodels.py
```diff
@@ -254,6 +261,8 @@ def set_state(self, new_state: EventState, user: str = "monitor"):
         old_state, self.state = self.state, new_state
         if (old_state, new_state) == (EventState.EMBRYONIC, EventState.OPEN):
             self.opened = now()
+        if old_state == EventState.CLOSED:
+            raise ClosedEventError
```
stveit, line 265: I would  put the ClosedEventError check further upp, above `old_state, self.state = self.state, new_state`. It looks like `self.state` gets set to `new_state`, and then afterwards you raise the ClosedEventError.

stveit, line 265: I would  put the ClosedEventError check further upp, above `old_state, self.state = self.state, new_state`. It looks like `self.state` gets set to `new_state`, and then afterwards you raise the ClosedEventError. I guess the test works since this doesnt get Commited (?), but would feel cleaner to just not update `self.state` at all

stveit (approved):
Looks fine to me

johannaengland:
I would like to wait for a review from @lunkwill42 before merging, mainly for the design decision if this is enough to close #188.

lunkwill42 (approved):
Looks a-ok to me! Thanks!

Issue #182, opened 2024-03-13
Add a task that automatically deleted closed events after a certain time


johannaengland:
In Zino 1.0 both regular events as well as maintenance events are deleted after being closed for a certain time. We should also have a task for this.

lunkwill42:
Correction: There is no difference between "maintenance" events and "regular" events.  PM events may be "synthetic", but they are seen as any other event in the event list.

runborg:
When a case is deleted from the database its properties should also be exported to a file so its inspectable in retrospect. 
This again could be used to generate statistics on failed connections and so on. 

This is by zino1 done by exporting a file in a directory structure consistent of `[zino-root]/old-events/[year-month]/[day]/` with the caseid as a filename.

The format of the existing files are in typical tcl sourceable format as in this example:  
```
[09:20:20] runarb@hugin.uninett.no /a/traffic/zino/old-events/2024-03
--> $ cat 16/170798
global EventAttrs_170798
set EventAttrs_170798(flaps) "1"
set EventAttrs_170798(history) "{1710519023 state change embryonic -> open (monitor)} {1710520485 he\n Nye logiske interfacer som oppst�r ifbm.\n l3vpn konfigurasjon for Havforskningsinstituttet.\n } {1710520485 state change open -> closed (he)}"
set EventAttrs_170798(id) "170798"
set EventAttrs_170798(ifindex) "738"
set EventAttrs_170798(log) "{1710519023 bergen-gw3: intf \"ge-100/0/33.16386\" ix 738 linkUp}"
set EventAttrs_170798(opened) "1710519023"
set EventAttrs_170798(polladdr) "128.39.0.51"
set EventAttrs_170798(port) "ge-100/0/33.16386"
set EventAttrs_170798(portstate) "up"
set EventAttrs_170798(priority) "100"
set EventAttrs_170798(router) "bergen-gw3"
set EventAttrs_170798(state) "closed"
set EventAttrs_170798(type) "portstate"
set EventAttrs_170798(updated) "1710519023"
```

The format of these files could be discussed, and i think it certainly could be exported as eg. yaml or json

PR #203, opened 2024-04-02
Add a task that automatically deletes closed events after a certain time

johannaengland:
First part to fixing #182.

I modeled it pretty closely to what happens in Zino1 (as far as I understood it):
- remove an event from the index the moment it is closed
- completely delete a closed event 8 hours after it has been closed

Dumping the event before it's deleted is happening in #204.

### DIFF
--- a/src/zino/events.py
+++ b/src/zino/events.py
@@ -57,8 +57,9 @@ class Events(BaseModel):
         """
         new_index: Dict[EventIndex, Event] = {}
         for event in self.events.values():
-            key = EventIndex(event.router, event.subindex, type(event))
-            new_index[key] = event
+            if event.state != EventState.CLOSED:
+                key = EventIndex(event.router, event.subindex, type(event))
+                new_index[key] = event
         self._events_by_index = new_index
 
     def get_or_create_event(
--- a/src/zino/events.py
+++ b/src/zino/events.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import timedelta
 from typing import Dict, NamedTuple, Optional, Protocol, Type, Union
 
 from pydantic.main import BaseModel
@@ -16,6 +17,8 @@ from zino.statemodels import (
 from zino.time import now
 
 _log = logging.getLogger(__name__)
+
+EVENT_EXPIRY = timedelta(hours=8)
 
 
 class EventIndex(NamedTuple):
@@ -143,6 +146,28 @@ class Events(BaseModel):
             self._events_by_index[index] = event
 
         self._call_observers_for(new_event=event, old_event=old_event)
+
+    def _delete(self, event: Event):
+        """Removes a closed event from the events dict and notifies all observers"""
+        if event.state != EventState.CLOSED:
+            return
+
+        index = EventIndex(event.router, event.subindex, type(event))
+        if self._events_by_index.get(index) and event.id == self._events_by_index[index].id:
+            _log.info("Closed event %s was still in event index, removing it now", event.id)
+            del self._events_by_index[index]
+        try:
+            del self.events[event.id]
+        except KeyError:
+            pass
+        self._call_observers_for(new_event=event)
+
+    def delete_expired_events(self):
+        """Deletes all events that have been closed for a certain time"""
+        event_list = list(self.events.values())
+        for event in event_list:
+            if event.state == EventState.CLOSED and now() > (event.updated + EVENT_EXPIRY):
+                self._delete(event)
 
     def add_event_observer(self, observer: EventObserver):
         """Adds an observer function that will be called any time an event is committed"""
--- a/src/zino/events.py
+++ b/src/zino/events.py
@@ -13,6 +13,7 @@ from zino.statemodels import (
     ReachabilityEvent,
     SubIndex,
 )
+from zino.time import now
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tests/events_test.py
```diff
@@ -104,6 +106,23 @@ def test_commit_should_open_embryonic_event(self):
...
+        events.commit(event)
+        assert not events.get(*identifiers)
+
+    def test_commit_should_set_updated_when_closing_event(self):
+        identifiers = "foobar", None, ReachabilityEvent
+        events = Events()
+        event = events.get_or_create_event(*identifiers)
+        event.set_state(EventState.CLOSED)
+        events.commit(event)
+        assert (now() - timedelta(minutes=1)) < event.updated < (now())
```
lunkwill42, line 124: I'm left wondering: What value did `updated` have before the commit? Are they different? Should the test perhaps confirm that too?

src/zino/events.py
```diff
@@ -144,6 +147,21 @@ def commit(self, event: Event, user: str = "monitor"):
...
+    def delete(self, event: Event):
+        """Removes an event from the events dict and notifies all observers"""
+        try:
+            del self.events[event.id]
+        except KeyError:
+            pass
+        self._call_observers_for(new_event=event)
+
+    def delete_closed_events(self):
+        """Deletes all events that have been closed for at least 8 hours"""
```
lunkwill42, line 159: No need to mention the number 8 hours in the docstring here.  It's defined by a constant.  Now, if you change the constant, you also have to remember to change the docstring...

lunkwill42 (changes requested):
Looks pretty good so far.

I think `Events._rebuild_indexes()` also needs to be updated.  It can be called to rebuild the entire index - this usually happens only when an `Events` object is being de-serialized from disk when Zino starts up.  It should also skip closed events.

> The only thing I am uncertain about it how to correctly notify the observers,

Do they need to be called at all?  The event closure was already sent as a notification.  Does Zino 1 send notifications when closed events eventually expire completely?

johannaengland:
> > The only thing I am uncertain about it how to correctly notify the observers,
> 
> Do they need to be called at all? The event closure was already sent as a notification. Does Zino 1 send notifications when closed events eventually expire completely?

Yes, Zino1 does call the observers. You can have a look at the function `notifyScavenge` in `server.tcl`.

tests/events_test.py
johannaengland: I am uncertain if by commit you mean the commit 77f2f5bbf37f6b1dcbbeac5215d141094e9aa9ec or you mean `events.commit(event)`.

If you mean the first we only so far set the `updated` when adding a log entry to an event.

If you mean the second then yes, `updated` would be `None` beforehand. 

But in either way, yes, if would be good to confirm that is actually changes.

src/zino/events.py
```diff
@@ -144,6 +147,21 @@ def commit(self, event: Event, user: str = "monitor"):
 
         self._call_observers_for(new_event=event, old_event=old_event)
 
+    def delete(self, event: Event):
+        """Removes an event from the events dict and notifies all observers"""
+        try:
+            del self.events[event.id]
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
johannaengland: Very good points!

tests/events_test.py
lunkwill42: I meant `event.commit()`, yes.  And I think it might still be good to have it in the test, in case something changes `event.updated` before the commit, down the road.

Maybe the test should just set `event.updated` to a known value, and then after the commit assert that `event.updated` has indeed been changed to a newer timestamp.

src/zino/events.py
```diff
@@ -147,20 +148,25 @@ def commit(self, event: Event, user: str = "monitor"):
...
 
-    def delete(self, event: Event):
-        """Removes an event from the events dict and notifies all observers"""
-        try:
-            del self.events[event.id]
-        except KeyError:
-            pass
+    def _delete(self, event: Event):
+        """Removes a closed event from the events dict and notifies all observers"""
+        if event.state is EventState.CLOSED:
```
lunkwill42, line 153: I would use an early return here instead, so there would be no need for an extra indent for the rest of the method.  I.e.

```suggestion
        if event.state != EventState.CLOSED:
           return
```

Other than that, be careful to distinguish properly between `=` and `is`.  `is` is an identity test, not an equality test.  It's best to use it only when you're sure you want to match objects by identity, not by value (like `is None`, since there is always just a single `None` object in a Python runtime)

src/zino/events.py
```diff
@@ -147,20 +148,25 @@ def commit(self, event: Event, user: str = "monitor"):
...
+                del self._events_by_index[index]
+            try:
+                del self.events[event.id]
+            except KeyError:
+                pass
         self._call_observers_for(new_event=event)
 
     def delete_closed_events(self):
-        """Deletes all events that have been closed for at least 8 hours"""
+        """Deletes all events that have been closed for a certain time"""
```
lunkwill42, line 165: Actually, come to think of it, a better name for this is `delete_expired_events`.  This doesn't delete closed events, per se, it only deletes closed events that have *exipred*.

src/zino/zino.py
```diff
@@ -51,7 +51,7 @@ def init_event_loop(args: argparse.Namespace):
         func=state.state.events.delete_closed_events,
         trigger="interval",
         minutes=30,
-        next_run_time=datetime.now(),
+        next_run_time=datetime.now() + timedelta(minutes=30),
```
lunkwill42, line 54: I believe you can drop this line completely. The first iteration of this job will always be scheduled to 30 minutes in the future unless you explicitly set a `next_run_time`.  This is why we set `next_run_time` in the other jobs, because we *do* want their first iteration to run immediately.

johannaengland:
I made the mistake of not running the tests before pushing this commit. The problem is that we don't have access to the state since `build_and_send_notifications` and `build_notifications` are class methods. This needs more concentration than I currently have on a Friday afternoon, so this is a problem for future me.

lunkwill42 (approved):
Good enough for me :)

PR #204, opened 2024-04-03
Dump event content to file on deletion

### LINKED ISSUE #182: Add a task that automatically deleted closed events after a certain time

johannaengland:
Closes #182. Dependent on #203. 

The format of the way the event is dumped can be discussed, I did it the easiest way for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist/
 
 # Generated files
 src/zino/version.py
+old-events/
 .coverage
 *.pyc
 *.egg-info/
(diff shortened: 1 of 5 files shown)


lunkwill42 (commented):
For me, the format is just as expected. This is a big part of the reason we have employed Pydantic in Zino 2 :)

I would approve, but the dependent PR needs to be fixed and merged first!

lunkwill42 (approved):

PR #217, opened 2024-05-21
Delete some tests just for workflow testing

lunkwill42:
We want to observe that the reduction in code coverage is properly measured and detected by CodeCov in our workflows.

**This PR should never be merged to master**


lunkwill42:
@johannaengland may look at this later (in a fork or throwaway repo).  Changes on master have been reverted until we have a better solution (so we can still get work done on this repo)

PR #221, opened 2024-05-23
Keep GitHub actions updated with Dependabot

johannaengland:
It would be good to not manually have to keep on top of all action updates.

Reference: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


lunkwill42 (approved):
Good idea! :)

PR #227, opened 2024-05-24
Rework trap observers to full classes with provisions auto-subscribing

lunkwill42:
Now that we're on the brink of implementing more actual trap handlers, it's time to slightly rethink the original alpha implementation of the trap observer protocol.

This reworks the protocol from being simple functions to being slightly more involved classes (which can inherit directly from the protocol to remain somewhat simple).  I realized that many of the observers will need access to the running `ZinoState` as well as the running event loop (which handlers already had an argument for).  It makes no sense to keep adding arguments to trap handler functions for everything not all of them need.  Instead, these more permanent values can be fed to the trap observers inherited constructor, and the handler function itself needs only concern itself with the incoming trap message argument.

This makes for even stronger separation of concerns.

This PR moves the list of ignored traps to a separate `TrapObserver` subclass, leaving it cleanly out of the `TrapReceiver` implementation itself, and serving as a sort of template of how the remaining observers can be implemented.

### DIFF
--- a/src/zino/trapobservers/ignored_traps.py
+++ b/src/zino/trapobservers/ignored_traps.py
@@ -1,0 +1,23 @@
+"""This module implements a trap observer to ignore spammy traps"""
+import logging
+from typing import Optional, Set
+
+from zino.trapd import TrapMessage, TrapObserver, TrapType
+
+_logger = logging.getLogger(__name__)
+
+
+class IgnoreTraps(TrapObserver):
+    """Completely ignores spammy traps observed in Uninett, that we don't care about"""
+
+    WANTED_TRAPS: Set[TrapType] = {
+        ("BGP4-MIB", "bgpBackwardTransition"),
+        ("BGP4-MIB", "bgpBackwardTransNotification"),
+        ("BGP4-V2-MIB-JUNIPER", "jnxBgpM2BackwardTransition"),
+        ("SNMPv2-MIB", "authenticationFailure"),
+        ("CISCOTRAP-MIB", "tcpConnectionClose"),
+        ("BGP4-V2-MIB-JUNIPER", "jnxBgpM2BackwardTransition"),
+    }
+
+    def handle_trap(self, trap: TrapMessage) -> Optional[bool]:
+        return False  # Stop processing here!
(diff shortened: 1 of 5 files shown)


src/zino/trapobservers/ignored_traps.py
```diff
@@ -0,0 +1,23 @@
...
+class IgnoreTraps(TrapObserver):
+    """Completely ignores spammy traps observed in Uninett, that we don't care about"""
+
+    WANTED_TRAPS: Set[TrapType] = {
+        ("BGP4-MIB", "bgpBackwardTransition"),
+        ("BGP4-MIB", "bgpBackwardTransNotification"),
+        ("BGP4-V2-MIB-JUNIPER", "jnxBgpM2BackwardTransition"),
+        ("SNMPv2-MIB", "authenticationFailure"),
+        ("CISCOTRAP-MIB", "tcpConnectionClose"),
+        ("BGP4-V2-MIB-JUNIPER", "jnxBgpM2BackwardTransition"),
```
johannaengland, line 19: Duplicate line

src/zino/zino.py
```diff
@@ -45,6 +45,9 @@ def init_event_loop(args: argparse.Namespace, loop: Optional[AbstractEventLoop]
         trap_receiver.add_community("public")
         trap_receiver.add_community("secret")
 
+        # ensure observers are loaded
+        from zino.trapobservers import ignored_traps  # noqa
```
johannaengland, line 49: Why do we have to do this here? And will we have to import all trap observers like this here?

src/zino/trapobservers/ignored_traps.py
lunkwill42: Ah, that's funny.  This was just moved from the original implementation, where no-one noticed the duplication :laughing:

src/zino/zino.py
lunkwill42: It doesn't have to happen here, it can be a normal module-level import.  But yes.  Every trap observer we want to include in the daemon would have to be imported explicitly.  If not, we also have to make a mechanism that finds all the modules and loads them - and for some things I like to just defer to the Zen of Python: _"Explicit is better than implicit"_ :grin:

johannaengland (approved):

johannaengland (approved):

PR #229, opened 2024-05-27
Add logging of `coldStart` and `warmStart` traps

### LINKED ISSUE #226: Implement `SNMPv2-MIB::coldStart` and `SNMPv2-MIB::warmStart` trap loggers
Start with this just to see how trap observers in general should be built and loaded by the codebase.

Zino 1 logs these very simply as `%s: coldStart` and `%s: warmStart`, where `%s` is the router name.

lunkwill42:
Closes #226 

Depends on #227

### DIFF
--- a/src/zino/trapobservers/logged_traps.py
+++ b/src/zino/trapobservers/logged_traps.py
@@ -1,0 +1,17 @@
+"""This module implements custom logging of certain SNMP traps, but takes no other action"""
+import logging
+from typing import Optional
+
+from zino.trapd import TrapMessage, TrapObserver
+
+_logger = logging.getLogger(__name__)
+
+
+class RestartTrapLogger(TrapObserver):
+    WANTED_TRAPS = {
+        ("SNMPv2-MIB", "coldStart"),
+        ("SNMPv2-MIB", "warmStart"),
+    }
+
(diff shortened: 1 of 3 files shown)


src/zino/trapobservers/logged_traps.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class RestartTrapLogger(TrapObserver):
+    WANTED_TRAPS = {
+        ("SNMPv2-MIB", "coldStart"),
+        ("SNMPv2-MIB", "warmStart"),
+    }
+
+    def handle_trap(self, trap: TrapMessage) -> Optional[bool]:
+        _logger.info("%s: %s", trap.agent.device.name, trap.name)
```
johannaengland, line 17: Shouldn't this also return `False`? Since that is the only thing that is done with these traps and therefore we can stop processing this trap afterwards?

lunkwill42: The return value is implicitly False.

The annotated return type is `Optional[bool]`.  If there is no return statement, `None` is Python's default return value.  `bool(None) == False`.

Anyway, `explicit is better than implicit`, so we could still add an explicit return :)

johannaengland (approved):
Beautiful

PR #230, opened 2024-05-28
Factor out codecov upload

### LINKED ISSUE #216: Make "upload codecov results" action a seperate action triggered by "test" action
We have figured out while working on Uninett/nav#2914 and Uninett/nav#2922 that when a workflow is triggered by another workflow it is run in a secure environment on the master branch and will have access to secrets of that repository.

This might be a good solution to our recent problem of codecov uploads failing since most of the developers work on their own forks, where the codecov action does not have access to the `CODECOV_TOKEN` and is therefore limited to tokenless uploading.

Which is why when there are many actions running at the same time will lead to the action failing since the limit for tokenless uploading is reached and it very annoying to have to run the action again.

johannaengland:
Closes #216.

Since upgrading the codecov action to version 4 uploading coverage reports without a token (so when creating a PR from a fork) is severely rate limited, which resulted often enough in our tests seemingly failing, even though it was only the coverage upload failing. (Reference: https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov)

Workflows that are triggered by other workflows are always run within the context of the main repository, instead of within the fork (which workflows triggered by pull requests for example can be). This means that that way we can have access to the secret `CODECOV_TOKEN` and avoid the rate limitations.

Since we already upload our coverage results as artifacts in the `Upload test reports` step of the `tests` workflow, we decided to move the uploading coverage step to the same workflow where we publish the test results: `publish-test-results`. This workflow is triggered by the completion of the `tests` workflow, which means it is run within the context of the main repository. 

Moving the uploading coverage step to the `publish-test-results` worked great in the way that we have access to the `CODECOV_TOKEN`, but due to the context switch codecov could not properly tell anymore which branch, commit and pull request that coverage information belongs to. 

That is why we override the pull request, commit and branch information. 

Getting the associated pull request number is quite a bit more complicated and it involves saving that number in a file, then uploading it as an artifact, then downloading that artifact and after validation setting that as the pull request number for codecov. 

I have tested this with four different scenarios:
1. Push against main, no associated pull request number ([tests](https://github.com/johannaengland/testing/actions/runs/9268837792), [upload](https://github.com/johannaengland/testing/actions/runs/9268845086))
2. Pull request with passing tests ([tests](https://github.com/johannaengland/testing/actions/runs/9268749420), [upload](https://github.com/johannaengland/testing/actions/runs/9268760548))
3. Pull request with failing tests ([tests](https://github.com/johannaengland/testing/actions/runs/9268759054), [upload](https://github.com/johannaengland/testing/actions/runs/9268768227))
4. Pull request that overrides the variable written to the file that will be uploaded ([tests](https://github.com/johannaengland/testing/actions/runs/9268765102), [upload](https://github.com/johannaengland/testing/actions/runs/9268778645))

This also upgrades various actions used in these workflows. 
Reference for why we need to rename the artifacts after upgrading the upload artifacts action: https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,10 +25,10 @@ jobs:
 
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         id: cache
         with:
           path: ~/.cache/pip
@@ -38,7 +38,7 @@ jobs:
 
 
       - name: "Set up Python ${{ matrix.python-version }}"
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,13 +51,6 @@ jobs:
       - name: Test with tox
         run: tox
 
-      - name: "Upload coverage to Codecov"
-        if: github.repository_owner == 'Uninett'
-        uses: codecov/codecov-action@v4
-        with:
-          fail_ci_if_error: true
-          token: ${{ secrets.CODECOV_TOKEN }}
-
       - name: Upload test reports (${{ matrix.python-version }})
         if: always()
         uses: actions/upload-artifact@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,8 +60,8 @@ jobs:
 
       - name: Upload test reports (${{ matrix.python-version }})
         if: always()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: reports
+          name: reports-${{ matrix.python-version }}
           path: |
             reports/**/*
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,3 +58,20 @@ jobs:
           name: reports-${{ matrix.python-version }}
           path: |
             reports/**/*
+
+  upload-pr-number:
+    name: Save PR number in artifact
+    runs-on: ubuntu-latest
+    if: ${{ github.event.number && always() }}
+    env:
+      PR_NUMBER: ${{ github.event.number }}
+    steps:
+      - name: Make PR number file
+        run: |
+          mkdir -p ./pr
+          echo $PR_NUMBER > ./pr/pr_number
+      - name: Upload PR number file
+        uses: actions/upload-artifact@v4
+        with:
+          name: pr_number
+          path: pr/
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -30,7 +30,7 @@ jobs:
            done
 
       - name: "Publish test results"
-        uses: EnricoMi/publish-unit-test-result-action/composite@v1
+        uses: EnricoMi/publish-unit-test-result-action@v2
         with:
           commit: ${{ github.event.workflow_run.head_sha }}
           check_name: "Test results"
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -51,17 +51,27 @@ jobs:
...
-          fail_ci_if_error: true
-          token: ${{ secrets.CODECOV_TOKEN }}
-
       - name: Upload test reports (${{ matrix.python-version }})
         if: always()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: reports
+          name: reports-${{ matrix.python-version }}
```
lunkwill42, line 58: I think that this step could potentially also be moved outside the matrix.  It seems the old version uploaded the same artifact several times, but the latest upload would include the test reports generated by all matrix dimensions (I assume the all work on the same file system) - so why would we need multiple artifacts if all the report files are available in the same directory structure after the matrix finishes?

lunkwill42 (approved):
Looks good to me. I'm sad there's no better way to communicate the PR number to the second workflow, but I guess that's GitHub actions for ya :-P

.github/workflows/tests.yml
johannaengland: No, the old version also uploads an artifact for each Python version, which you can see for example in [this workflow](https://github.com/Uninett/zino/actions/runs/9268673231/job/25497813462) - the difference is just that they are merged into one folder
But yes, it is a good idea to move the uploading step outside to upload all test results at the same time

lunkwill42: I think you misunderstood me.  The old version clearly uploads multiple times, once for each Python version, which I wrote. But the artifact is named the same on every upload, so it's really the same artifact.

The question is whether the resulting artifact contains all test reports because of the old uploader action's magic artifact-merging functionality, or because all matrix dimensions share a common workspace directory?

If all matrix dimensions share a common workspace directory, the contents of the `report/` directory will grow for each matrix dimension that runs, meaning that each artifact step will just package more and more data, regardless of any magic artifact-merging.

If the matrix dimensions operate on isolated workspace directories, then we were dependent on the artifact-merging, and using the new uploader means we *do* have to produce multiple artifacts.  If the workspace is shared, we can get away with a single artifact produced at the end of the test run.

I genuinely don't know which is the case with workflow matrixes, and if you also don't, we should check the docs :)

johannaengland: Ah, I see, yes, the old uploader is able to merge these artifacts. I read somewhere in my research, that previously these artifact were mutable, but with the new action version they are not anymore. 

I work under the assumption that all matrix dimensions share a common workspace directory, I am currently in the process of testing if that is the case and that we can that way upload the report artifact after all tests have run.

johannaengland: From the release announcement of the artifacts action: "Using upload-artifact@v4 ensures artifacts are immutable"

johannaengland:
Also just to have it documented somewhere: 
In a discussion on how to get the pull request number since it is not easily accessible when running in a workflow triggered by `workflow_run` it was mentioned to check how the action [publish-test-results](https://github.com/EnricoMi/publish-unit-test-result-action) handles it.
I figured out it is how we ended up doing it: uploading an artifact that has all information about the event that triggered the workflow that will trigger the workflow that runs the `publish-test-results` action. That artifact is then used later to determine all the details like pull request number, etc. 
The upload can be seen here: https://github.com/EnricoMi/publish-unit-test-result-action/blob/master/.github/workflows/ci-cd.yml#L175

.github/workflows/tests.yml
lunkwill42: I suspected that could be the case.  But you didnt't change anything, and I didn't ask for this PR to change anything (it was a potential future issue to work on) - I approved the PR, so no need to re-request and wait for another review :smile:

lunkwill42 (approved):
Still approved...

PR #233, opened 2024-05-29
Implement link trap handling (without flapping algorithm)

### LINKED ISSUE #232: Implement handling of link state traps (`IF-MIB::linkUp` and `IF-MIB::linkDown`)
This should follow Zino 1 closely.   Flap handling could potentially be a separate PR to keep this as small as possible.

lunkwill42:
Partially closes #232.  Makes provisions for handling flapping, but I'm trying to postpone it to a separate PR.

This reworks `TrapObserver` and the `PortStateEvent` model as needed.  So far, it lacks provisions to actually schedule poll jobs, as that is currently built as private methods of the linkstatetask.  I'm thinking this PR is already large enough for a review and for some simple functionality, so I suggest a followup PR to deal also with rescheduling, which will require more restructuring.

### DIFF
--- a/src/zino/trapd.py
+++ b/src/zino/trapd.py
@@ -4,7 +4,7 @@ import asyncio
 import logging
 from dataclasses import dataclass, field
 from ipaddress import ip_address
-from typing import Any, List, NamedTuple, Optional, Set
+from typing import Any, Dict, List, NamedTuple, Optional, Set
 
 from pysnmp.carrier.asyncio.dgram import udp
 from pysnmp.entity import config
@@ -12,9 +12,10 @@ from pysnmp.entity.rfc3413 import ntfrcv
 from pysnmp.proto.rfc1902 import ObjectName
 from pysnmp.smi import view
 
+import zino.state
+from zino.config.models import PollDevice
 from zino.oid import OID
 from zino.snmp import _mib_value_to_python, get_new_snmp_engine
-from zino.state import ZinoState
 from zino.statemodels import DeviceState, IPAddress
 
 TrapType = tuple[str, str]  # A mib name and a corresponding trap symbolic name
@@ -64,8 +65,18 @@ class TrapObserver:
 
     WANTED_TRAPS: Set[TrapType] = set()
 
-    def __init__(self, state: ZinoState, loop: Optional[asyncio.AbstractEventLoop] = None):
+    def __init__(
+        self,
+        state: zino.state.ZinoState,
+        polldevs: Optional[Dict[str, PollDevice]] = None,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
+    ):
+        """Initializes a new trap observer with a reference to Zino's state.
+
+        Optionally also receives a reference to the current polldevs configuration and the running event loop.
+        """
         self.state = state
+        self.polldevs: Dict[str, PollDevice] = polldevs or zino.state.polldevs
         self.loop = loop if loop else asyncio.get_event_loop()
 
     def handle_trap(self, trap: TrapMessage) -> Optional[bool]:
@@ -91,13 +102,15 @@ class TrapReceiver:
         address: str = "0.0.0.0",
         port: int = 162,
         loop: Optional[asyncio.AbstractEventLoop] = None,
-        state: Optional[ZinoState] = None,
+        state: Optional[zino.state.ZinoState] = None,
+        polldevs: Optional[Dict[str, PollDevice]] = None,
     ):
         self.transport: udp.UdpTransport = None
         self.address = address
         self.port = port
         self.loop = loop if loop else asyncio.get_event_loop()
-        self.state = state or ZinoState()
+        self.state = state or zino.state.ZinoState()
+        self.polldevs = polldevs if polldevs is not None else {}
         self.snmp_engine = get_new_snmp_engine()
         self._communities = set()
         self._observers: dict[TrapType, List[TrapObserver]] = {}
@@ -112,7 +125,7 @@ class TrapReceiver:
                 continue
             else:
                 self._auto_subscribed_observers.add(observer_class)
-            observer_instance = observer_class(self.state, self.loop)
+            observer_instance = observer_class(state=self.state, polldevs=self.polldevs, loop=self.loop)
             self.observe(observer_instance, *observer_instance.WANTED_TRAPS)
 
     def observe(self, subscriber: TrapObserver, *trap_types: List[TrapType]):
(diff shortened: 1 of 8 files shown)


src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # If we don't know that the device restarted recently, do not ignore
+        if not device.boot_time:
+            _logger.info("Oops! Do not know restart time for %s", device.name)
+            return False
+
+        if not port.state:
+            port.state = InterfaceState.UNKNOWN
+
+        if now - device.boot_time > timedelta(minutes=5):
```
stveit, line 140: This magic 5 minutes could be a static value since its used a couple places and might want to configure in the future

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,181 @@
...
+            # Do not ignore traps indicating different state than known
+            current_state = InterfaceState.UP if is_up else InterfaceState.DOWN
+            if port.state != current_state:
+                return False
+
+            # Do not ignore successive traps within a 5 minute interval indicating same state as previously known
+            index = (device.name, port.ifindex)
+            if last_same_trap := self._last_same_trap.get(index):
+                if now - last_same_trap < timedelta(minutes=5):
```
stveit, line 149: This magic 5 minutes could be a static value since its used a couple places and might want to configure in the future
(assuming this value is related to the other 5 minute timedelta used above)

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,181 @@
...
+                    self._last_same_trap[index] = now
+                    return False
+            self._last_same_trap[index] = now
+
+        # Ignoring traps the first 5 minutes after restart
```
stveit, line 154: 5 minute referenced again

stveit (commented):
Surface level review

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+        return True
+
+    def update_interface_flapping_score(self, index: Tuple[str, int]) -> bool:
+        """Updates the running flapping score for a given port"""
+        pass  # stub implementation, see Zino 1 `proc intfFlap`
+
+    def is_interface_flapping(self, index: Tuple[str, int]) -> bool:
+        """Determines if a given port is flapping"""
+        return False  # stub implementation, see Zino 1 `proc flapping`
```
stveit, line 171: I know these are just placeholders, but could still have consistent return values

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,181 @@
...
+        # TODO: The trap *might* contain an ifDescr value.  If present, Zino uses that for trap processing.
+        #  Otherwise, it fetches ifDescr from its own state and uses that for trap processing.  Either way,
+        #  there seems to be some redundancy. We should document why, or change the behavior in Zino 2
+
+        # The legacy Zino also looks for `locIfDescr` in the trap at this point, but this is an ancient
+        # Cisco-specific variable we no longer see.  It might have been replaced by `cieIfOperStatusCause`,
+        # but we have no more Cisco devices to test on, so this has been deliberately left out.
+
+        is_up = trap.name == "linkUp"
+        self.handle_link_transition(trap.agent.device, port, is_up)
```
stveit, line 43: Port is set to `None` here if `ifindex > 0` and then put as argument to `self.handle_link_transition` which does not accept `None` as a value for `port`. The `port` value is then further used in other functions that only accept the `Port` type. 

I imagine the correct thing to do is update these typehints to `Optional[Port]`

src/zino/trapobservers/link_traps.py
lunkwill42: I assume you mean it could be a *constant*... fixed (although I'm not sure what to call this constant)

src/zino/trapd.py
```diff
@@ -91,13 +102,15 @@ def __init__(
...
+        state: Optional[zino.state.ZinoState] = None,
+        polldevs: Optional[Dict[str, PollDevice]] = None,
     ):
         self.transport: udp.UdpTransport = None
         self.address = address
         self.port = port
         self.loop = loop if loop else asyncio.get_event_loop()
-        self.state = state or ZinoState()
+        self.state = state or zino.state.ZinoState()
+        self.polldevs = polldevs if polldevs is not None else {}
```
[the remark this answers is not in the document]
lunkwill42: What default value?  You mean change the function argument default value from `None` to `{}`?   That would be an anti-pattern in Python.  The default value will be defined and created only once and then shared between all instances where the value is omitted by the caller.

src/zino/trapobservers/link_traps.py
lunkwill42: Fixed

src/zino/trapobservers/link_traps.py
lunkwill42: I followed the original implementation a bit too closely here, methinks.  

It's not really clear *what* does happen in the original when the trap doesn't contain an `ifIndex` value, except that it's set internally to `-1` and then used for lookups.  But of course, it won't find such an interface in its internal data structures.

I'm thinking the best strategy here is to drop the trap immediately, since it's of no use.

lunkwill42: Added a commit to this effect

src/zino/trapd.py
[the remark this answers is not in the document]
stveit: I did not know that, thats a hilarious effect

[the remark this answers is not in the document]
lunkwill42: It's a classic Python trap :grin: 

```pycon
>>> def foo(arg=[]):
...     arg.append("foo")
...     print(arg)
...
>>> foo()
['foo']
>>> foo()
['foo', 'foo']
>>> foo()
['foo', 'foo', 'foo']
>>> foo(['bar'])
['bar', 'foo']
>>>
```

johannaengland (approved):
Just a small documentation comment within the tests, otherwise looks good

stveit (approved):

tests/trapobservers/link_traps_test.py
```diff
@@ -0,0 +1,61 @@
...
+    ):
+        assert not state_with_localhost_with_port.events.get(
+            "localhost", 1, PortStateEvent
+        ), "initial state should be empty"
+
+        observer = LinkTrapObserver(
+            state=localhost_receiver.state, polldevs=localhost_receiver.polldevs, loop=localhost_receiver.loop
+        )
+        localhost_receiver.observe(observer, *LinkTrapObserver.WANTED_TRAPS)
+        await trapd_test.send_trap_externally(OID_LINKDOWN, OID_IFINDEX, "i", "1", OID_IFOPERSTATUS, "i", "2")
```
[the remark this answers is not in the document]
lunkwill42: The comment would amount to `man snmptrap` - the esoteric arguments are only explained there.

PR #237, opened 2024-05-31
Exempt `raise NotImplementedError` from coverage

lunkwill42:
We just don't need it.


lunkwill42:
... and I don't think we need a towncrier file for this

stveit (approved):

johannaengland (approved):

PR #241, opened 2024-05-31
Schedule single interface poll jobs when link trap changes state

lunkwill42:
Continued implementation of #232 

Depends on #233 to be merged first.

### DIFF
--- a/src/zino/tasks/linkstatetask.py
+++ b/src/zino/tasks/linkstatetask.py
@@ -135,18 +135,20 @@ class LinkStateTask(Task):
         event.add_log(log)
         self.state.events.commit(event)
 
-        self._schedule_verification_of_single_port(port.ifindex)
+        self.schedule_verification_of_single_port(port.ifindex)
 
-    def _schedule_verification_of_single_port(self, ifindex: int):
-        in_two_minutes = datetime.datetime.now() + datetime.timedelta(minutes=2)
-        job_name = f"{self.device.name}-verify-{ifindex}-state"
+    def schedule_verification_of_single_port(
+        self, ifindex: int, deadline: datetime.timedelta = datetime.timedelta(minutes=2), reason: str = "verify"
+    ):
+        """Schedules a verification of a single port at a given time in the future"""
+        when = datetime.datetime.now() + deadline
+        job_name = f"{self.device.name}-{reason}-{ifindex}-state"
         self._scheduler.add_job(
             func=self.poll_single_interface,
             args=(ifindex,),
             trigger="date",
-            run_date=in_two_minutes,
+            run_date=when,
             name=job_name,
-            id=job_name,
         )
 
     def _get_or_create_port(self, ifindex: int):
--- a/src/zino/tasks/linkstatetask.py
+++ b/src/zino/tasks/linkstatetask.py
@@ -141,13 +141,13 @@ class LinkStateTask(Task):
         self, ifindex: int, deadline: datetime.timedelta = datetime.timedelta(minutes=2), reason: str = "verify"
     ):
         """Schedules a verification of a single port at a given time in the future"""
-        when = datetime.datetime.now() + deadline
+        verification_time = datetime.datetime.now() + deadline
         job_name = f"{self.device.name}-{reason}-{ifindex}-state"
         self._scheduler.add_job(
             func=self.poll_single_interface,
             args=(ifindex,),
             trigger="date",
-            run_date=when,
+            run_date=verification_time,
             name=job_name,
         )
 
--- a/src/zino/trapobservers/link_traps.py
+++ b/src/zino/trapobservers/link_traps.py
@@ -7,6 +7,7 @@ from typing import Optional, Tuple
 
 import zino.time
 from zino.statemodels import DeviceState, InterfaceState, Port, PortStateEvent
+from zino.tasks.linkstatetask import LinkStateTask
 from zino.trapd import TrapMessage, TrapObserver
 
 _logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+            # TODO: If there is internal flapping state, log it in the event and clear internal state
+            # TODO: Set final flapcount in event
+
+            msg = f'{device.name}: intf "{port.ifdescr}" ix {port.ifindex} link{new_state.capitalize()}'
+            if reason:
+                event.reason = reason
+                msg = f'{msg}, "{reason}"'
+            _logger.info(msg)
+            event.add_log(msg)
+            self.state.events.commit(event)
```
stveit, line 106: Could this event creation be moved into its own function to reduce the size of `handle_link_transition`?

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+        index = (device.name, port.ifindex)
+        self.update_interface_flapping_score(index)
+
+        new_state = InterfaceState.UP if is_up else InterfaceState.DOWN
+
+        if self.is_interface_flapping(index):
+            # TODO: if event doesn't exist, create it
+            # TODO: Record new number of flaps in event
+            # TODO: When flapcount modulo 100 is zero, log a message with flapping stats
+            pass
```
stveit, line 84: I imagine whatever will be in here should also be separated into its own function

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+            event.add_log(msg)
+            self.state.events.commit(event)
+
+            poll = LinkStateTask(device=polldev, state=self.state)
+            poll.schedule_verification_of_single_port(port.ifindex, deadline=timedelta(seconds=0), reason="trap-verify")
+            poll.schedule_verification_of_single_port(
+                port.ifindex, deadline=timedelta(minutes=2), reason="trap-verify-2"
+            )
+
+    def is_port_ignored_by_patterns(self, device: DeviceState, ifdescr: str) -> bool:
```
stveit, line 114: Add a docstring something like this:
```suggestion
    def is_port_ignored_by_patterns(self, device: DeviceState, ifdescr: str) -> bool:
        """Checks if a port should be ignored based on the watch and ignore patterns for the device"""
```

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def update_interface_flapping_score(self, index: Tuple[str, int]) -> bool:
+        """Updates the running flapping score for a given port"""
+        return False  # stub implementation, see Zino 1 `proc intfFlap`
+
+    def is_interface_flapping(self, index: Tuple[str, int]) -> bool:
+        """Determines if a given port is flapping"""
+        return False  # stub implementation, see Zino 1 `proc flapping`
+
+    def get_watch_pattern(self, device: DeviceState) -> Optional[str]:
```
stveit, line 184:
```suggestion
    def get_watch_pattern(self, device: DeviceState) -> Optional[str]:
        """Retrieves the watch pattern for a given device"""
```

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+        if device.name not in self.polldevs:
+            return None
+        return self.polldevs[device.name].watchpat
+
+    def get_ignore_pattern(self, device: DeviceState) -> Optional[str]:
```
stveit, line 189:
```suggestion
    def get_ignore_pattern(self, device: DeviceState) -> Optional[str]:
        """Retrieves the ignore pattern for a given device"""
```

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+        self._last_same_trap: dict[Tuple[str, int], datetime] = {}
+
+    def handle_trap(self, trap: TrapMessage) -> Optional[bool]:
+        _logger.debug("%s: %s (vars: %s)", trap.agent.device.name, trap.name, ", ".join(trap.variables))
+
+        if "ifIndex" in trap.variables:
+            ifindex = trap.variables.get("ifIndex").value
+        else:
+            _logger.warning("%s: %s trap contained no ifIndex value, ignoring", trap.agent.device.name, trap.name)
+            return False
```
stveit, line 34: `"ifIndex"`is used a few times, so maybe make this a constant?

Could do the same with `"linkUp"` and  `"linkDown"` for consistencies sake

src/zino/trapobservers/link_traps.py
```diff
@@ -0,0 +1,192 @@
...
+                msg = f'{msg}, "{reason}"'
+            _logger.info(msg)
+            event.add_log(msg)
+            self.state.events.commit(event)
+
+            poll = LinkStateTask(device=polldev, state=self.state)
+            poll.schedule_verification_of_single_port(port.ifindex, deadline=timedelta(seconds=0), reason="trap-verify")
+            poll.schedule_verification_of_single_port(
+                port.ifindex, deadline=timedelta(minutes=2), reason="trap-verify-2"
+            )
```
stveit, line 112: Even tho these timedeltas are only used once, it might still be good to make them constants (I have suggested constants quite a bit now so might get cluttered at the top of the file :stuck_out_tongue: )

stveit (changes requested):
Some non-logic related comments

src/zino/trapobservers/link_traps.py
lunkwill42: The two-minute deadline could do as a constant, but the 0 second one is just sort of a "workaround" for the fact that this isn't an async function, and we want to call an async function in a  "fire and forget" manner:   This function doesn't care about the results of the poll operation - it just needs to be scheduled be run *once*, as soon as the event loop resumes.

lunkwill42:
See, this is the problem with dependent PRs, @stveit - most of your comments are actually directed towards commits that really came from a PR you've already approved :)

I do agree there's a lot that can be cleaned up overall, but I'm trying to stick as close as possible to the original before any potential refactors.

src/zino/tasks/linkstatetask.py
```diff
@@ -135,18 +135,20 @@ def _make_or_update_state_event(self, port: Port, new_state: InterfaceState, las
...
-    def _schedule_verification_of_single_port(self, ifindex: int):
-        in_two_minutes = datetime.datetime.now() + datetime.timedelta(minutes=2)
-        job_name = f"{self.device.name}-verify-{ifindex}-state"
+        self.schedule_verification_of_single_port(port.ifindex)
+
+    def schedule_verification_of_single_port(
+        self, ifindex: int, deadline: datetime.timedelta = datetime.timedelta(minutes=2), reason: str = "verify"
+    ):
+        """Schedules a verification of a single port at a given time in the future"""
+        when = datetime.datetime.now() + deadline
```
johannaengland, line 144: I do not like the variable name `when` - what about something like `verification_time`?

lunkwill42: Done :)

johannaengland (approved):

PR #242, opened 2024-06-03
Do not try to publish test results/code coverage in fork

johannaengland:
When doing #230 I remove the condition `github.repository_owner == 'Uninett'` because I thought that was given, but I forgot, that the workflow `Run test suite` will also be triggered by pushes to the master branch of a fork, which will in turn trigger the `Publish test and coverage results` workflow. This workflow will then run on the master branch of the fork. 
And both uploading test results and code coverage will fail since the fork does not have the correct permissions/the code coverage token. Failing workflows can be seen here for [publish test results](https://github.com/johannaengland/zino/actions/runs/9345474386/job/25718806440) and [publish code coverage](https://github.com/england-johanna/testing/actions/runs/9345772140/job/25719278459).

So we do need to restrict this workflow to only run on the Uninett repository.

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -10,7 +10,7 @@ jobs:
   publish-test-results:
     name: "Publish test and coverage results"
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion != 'skipped'
+    if: github.event.workflow_run.conclusion != 'skipped' && github.repository_owner == 'Uninett'
 
     steps:
       - name: Download and Extract Artifacts


hmpf (changes requested):
Commit message: 

"Prevent publishing test results/code coverage in fork

This only works intermittently so cut down on the noise"

or something similar.

hmpf:
> Commit message:
> 
> "Prevent publishing test results/code coverage in fork
> 
> This only works intermittently so cut down on the noise"
> 
> or something similar.

.. or passive-aggressively mention something something about codecov behaving badly =)

johannaengland:
> Commit message:
> 
> "Prevent publishing test results/code coverage in fork
> 
> This only works intermittently so cut down on the noise"
> 
> or something similar.

But publishing test results/code coverage in fork never work, it is not a problem that is fail once in a while, it always fails, because it lacks the necessary permissions

But I can make the commit message a bit more detailed

hmpf:
"Avoid bothering codecov with code in forks

Codecov does not like forks."

Short, sweet, to the point and suitably #%&/(&%/(.

lunkwill42:
> "Avoid bothering codecov with code in forks
> 
> Codecov does not like forks."
> 
> Short, sweet, to the point and suitably #%&/(&%/(.

You guys are discussing and confusing two very different issues here:

1. Workflows that run in the upstream repo.
2. Workflows that run in a forked repo.


The issue with no. 1 is when a workflow is triggered by a PR from a forked repo. The workflow runs in the upstream repo, but in the context of the fork (so that things like upstream repo secrets are not available to the workflow).  CodeCov supports tokenless uploads for public upstream repos, so the secret isn't needed.  However, CodeCov also enforces rate limits for tokenless uploads, and GitHub has *many* anonymous requests all over for this API, so it becomes a problem for us.


The issue with no 2. is just when you push code to your forked repo, regardless of whether this is associated with a PR in an upstream repo or not: You likely did not configure CodeCov to accept data from the forked repo, so it's never going to work (unless you log in to CodeCov and explicitly add your forked repo as well).


It sounds to me like @johannaengland is discussing no. 2, while @hmpf is annoyed about no. 1.

hmpf (approved):
Remember to fit the first line of the commit message to the "This commit will.." pattern, easily done by squashing from github itself.

PR #249, opened 2024-06-10
Add generic Zino config file

### LINKED ISSUE #224: Add a generic Zino configuration file
Zino 1 features a config file for some generic settings (whereas the current Zino 2 beta hardcodes everything that isn't configurable through a command line option).

The original configuration file is (of course) plain Tcl, and looks something like this:

```tcl
set persist_file "save-state.tcl"
set eventid_file "eventid"
set persist_period 300

set old_events "old-events"

set pollfile "../conf/polldevs.cf"
set conf_check 60

set zino_secrets "secrets"
set authTacacs 1
set Tacping "/local/traffic/bin/tacping"
set Tachost "hegre.uninett.no"
set authClearTextPasswords 1
set OpenServerAccess 0
```

While some of these options aren't relevant for Zino 2 (because of features that have intentionally been dropped), several of them should be supported in a Zino 2 config file.

There is no reason to keep the original format (as Zino 2 is not written in Tcl): The config file is small and should be easy to port by hand.  In keeping with some of our other recent projects, I would suggest a TOML configuration file (`zino.toml`).  Existing  Zino 2 features (at the time of this writing) that should be configurable would be:

- `persist_file`, a.ka.a the file where persistent state is saved (defaults to `zino-state.json` atm).
- `persist_period`, how often the state should be saved (I assume when not otherwise rescheduled by Zino because of critical state changes)
- `eventid_file`: Not quite sure what this is, it should be checked
- `old_events`: The directory where expired events are saved
- `pollfile`: The path to a valid `polldevs.cf`.  This is currently configurable with a command line option in Zino 2.  At best, order of precedence here should be 1. command line option, 2. config file, 3. default value (`./polldevs.cf`)
- `conf_check`: Presumably how often `polldevs.cf` is re-read.  Maybe check whether `zino.cf` (and by extension, `zino.toml` is re-read on this interval as well)
- `zino_secrets`: The path to the file containing user accounts and secrets for the Zino API. Currently hardcoded in Zino 2 to just `secrets`.

Since we're redefining the config file, it's okay to group thing into TOML sections where applicable.

johannaengland:
Closes #224.

A few notes for discussion/information:
- as @runborg already commented in #224: `eventid_file` is not necessary, since it only keeps track of the last event id, which is something we already cover with `last_event_id`, which is dumped to `zino-state.json`

New questions that popped up during the fixing/writing tests:
- is it enough validation or does any other field need to be further validated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,18 @@ Zino will check `polldevs.cf` for changes on a scheduled interval while it's
 running, so any changes made while Zino is running should be picked up without
 requiring a restart of the process.
 
+### Configuring other settings
+
+Other settings can be also configured in a separate [TOML](https://toml.io/en/) file,
+which defaults to `zino.toml` in the current working directory, but a different file
+can be specified using the `--config-file` command line option.
+
+See the [zino.toml.example](./zino.toml.example) file for the settings that can be
+configured and their default values.
+
+Zino does not currently check `zino.toml` for changes on a scheduled interval while
+it's running, so Zino needs to be restarted for changes to take effect.
+
 ### Configuring API users
 
 Zino 2 reimplements the text-based (vaguely SMTP-esque) API protocol from Zino
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "pysnmplib",
     "pyasn1<0.5.0",
     "aiodns",
+    "tomli; python_version < '3.11'",
 ]
 dynamic = ["version"]
 
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -70,7 +70,7 @@ def init_event_loop(args: argparse.Namespace, loop: Optional[AbstractEventLoop] 
     scheduler.add_job(
         func=load_and_schedule_polldevs,
         trigger="interval",
-        args=(args.polldevs.name or state.config.polling.file,),
+        args=(state.config.polling.file,),
         minutes=state.config.polling.period,
         next_run_time=datetime.now(),
     )
@@ -188,7 +188,11 @@ def reschedule_dump_state(log_msg: str) -> None:
 def parse_args(arguments=None):
     parser = argparse.ArgumentParser(description="Zino is not OpenView")
     parser.add_argument(
-        "--polldevs", type=argparse.FileType("r"), metavar="PATH", default=None, help="Path to the pollfile"
+        "--polldevs",
+        type=argparse.FileType("r"),
+        metavar="PATH",
+        required=False,
+        help="Path to the pollfile",
     )
     parser.add_argument(
         "--config-file",
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -182,6 +182,13 @@ def parse_args(arguments=None):
         "--polldevs", type=argparse.FileType("r"), metavar="PATH", default="polldevs.cf", help="Path to polldevs.cf"
     )
     parser.add_argument(
+        "--config-file",
+        type=argparse.FileType("r"),
+        metavar="PATH",
+        required=False,
+        help="Path to zino configuration file",
+    )
+    parser.add_argument(
         "--debug", action="store_true", default=False, help="Set global log level to DEBUG. Very chatty!"
     )
     parser.add_argument("--stop-in", type=int, default=None, help="Stop zino after N seconds.", metavar="N")
@@ -199,6 +206,8 @@ def parse_args(arguments=None):
     args = parser.parse_args(args=arguments)
     if args.polldevs:
         args.polldevs.close()  # don't leave this temporary file descriptor open
+    if args.config_file:
+        args.config_file.close()  # don't leave this temporary file descriptor open
     return args
 
 
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -36,7 +36,7 @@ def main():
         level=logging.INFO if not args.debug else logging.DEBUG,
         format="%(asctime)s - %(levelname)s - %(name)s (%(threadName)s) - %(message)s",
     )
-    state.config = read_configuration(args.config_file)
+    state.config = read_configuration(args.config_file, args.polldevs)
     # Polldevs by command line argument will override config file entry
     if args.polldevs:
         state.config.polling.file = args.polldevs
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -16,6 +16,7 @@ import tzlocal
 from zino import state
 from zino.api.server import ZinoServer
 from zino.config import read_configuration
+from zino.config.models import validate_file_can_be_opened
 from zino.scheduler import get_scheduler, load_and_schedule_polldevs
 from zino.statemodels import Event
 from zino.trapd import TrapReceiver
@@ -39,6 +40,8 @@ def main():
     # Polldevs by command line argument will override config file entry
     if args.polldevs:
         state.config.polling.file = args.polldevs.name
+    else:
+        validate_file_can_be_opened(state.config.polling.file)
     state.state = state.ZinoState.load_state_from_file(state.config.persistence.file) or state.ZinoState()
     init_event_loop(args)
 
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -70,7 +70,7 @@ def init_event_loop(args: argparse.Namespace, loop: Optional[AbstractEventLoop] 
     scheduler.add_job(
         func=load_and_schedule_polldevs,
         trigger="interval",
-        args=(args.polldevs.name,),
+        args=(args.polldevs.name or state.config.polling.file,),
         minutes=1,
         next_run_time=datetime.now(),
     )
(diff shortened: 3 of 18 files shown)


hmpf:
> should we require a zino.toml file to exist? or is it optional and otherwise we simply use the default values?

If the zino.toml file is used to do the job of `polldevs.cf` there should be an error on startup if the polldevs-section is empty.

hmpf:
It can be nice to have an example zino.toml with all the defaults explicitly set and commented out. This is what postgres does.

runborg:
>>should we require a zino.toml file to exist? or is it optional and otherwise we simply use the default values?
i think everything in zino.toml should have it's default fallback in code-defaults. and those should be mandatory to have. 
But it would be great to have this file created with all default values documented so the user can uncomment and change it. 

>If the zino.toml file is used to do the job of polldevs.cf there should be an error on startup if the polldevs-section is empty.
I do not think it should be redundant options in polldevs.cf and zino.toml.. and everythin in polldevs.cf should be directly connected to one or more devices in the polling cycle.

hmpf:
Also, if multiple config files, all should use the same format unless one of them only contains one string.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "pydantic>=2",
     "pysnmplib",
     "pyasn1<0.5.0",
+    "tomli",
```
hmpf, line 28: Only needed for python < 3.11, python >= 3.11 comes with tomllib

tests/conftest.py
```diff
@@ -63,6 +63,22 @@ def polldevs_conf_with_no_routers(tmp_path):
...
+def zino_default_conf(tmp_path):
+    name = tmp_path / "zino.toml"
+    with open(name, "w") as conf:
+        conf.write(
+            """
+            old_events = "old-events"
+            zino_secrets = "secrets"
+            persist_file = "zino-state.json"
+            persist_period = 5
+            conf_check = 1
```
hmpf, line 76: .. and polldevs.cf

hmpf: This can be reused to generate the default toml-file on the fly, as a cli-command

zino.toml.example
runborg, line 1: I Think we should take a step back and identify all known possible configuration parameters, and use some time to get propper names and groups for the config file format.

example of possible config file options are things that you could enable with parameters today. eg. `--trap-port` and `--debug`-level

pyproject.toml
lunkwill42: This can be fixed with a modif

I guess this can be resolved with a `tomli  ; python_version < '3.11'` as a replacement, but I would assume an import guard is needed in the code as well, sinece the modules are named differently...

tests/conftest.py
lunkwill42: Not from the tests tree. It would have to be defined somewhere else.

I mean, it would be nice if Zino could generate the example config file from a command line option.  But really, all the values in `zino.toml` should have defaults, so that Zino actually *can* run without a config file present.

The minimum requirement for Zino to run is really the presence of a valid pollfile (`polldevs.cf`) with at least 1 device in it.  Without that, Zino has nothing to do and might as well exit.

src/zino/zino.py
```diff
@@ -35,7 +36,11 @@ def main():
         level=logging.INFO if not args.debug else logging.DEBUG,
         format="%(asctime)s - %(levelname)s - %(name)s (%(threadName)s) - %(message)s",
     )
-    state.state = state.ZinoState.load_state_from_file() or state.ZinoState()
+    state.config = read_configuration(args.config_file.name)
+    state.state = (
+        state.ZinoState.load_state_from_file(state.config.get("persistent_file", "zino-state.json"))
```
lunkwill42, line 41: Really, we shouldn't re-define the default value here, when it's already defined as `zino.state.STATE_FILENAME`

tests/zino_test.py
```diff
@@ -36,14 +36,16 @@ def test_zino_should_not_crash_right_away(polldevs_conf_with_no_routers):
             str(seconds_to_run_for),
             "--polldevs",
             str(polldevs_conf_with_no_routers),
+            "--config-file",
+            str(zino_default_conf),
```
lunkwill42, line 40: Why do we need to add a config-file option to this test?  The `zino.toml` file is not required to run Zino, but a `polldevs.cf` is (or there would be nothing for Zino to do).

Since a valid `polldevs.cf` is needed to run, adding this option here is just to ensure the Zino process has access to a valid pollfile during the test, so it can run a "normal" startup routine.

Testing that Zino actually picks up the settings from a valid `zino.toml` would be best as a separate test.  Also, maybe a test for what Zino should do in the case of an invalid `zino.toml` file.

lunkwill42 (changes requested):
You're on the right track here, but I believe it should still be possible to run Zino without having a config-file.  It should just run off its default values.  I might not have been too specific about this in the original issue.

A common strategy for this, used in other projects of ours, is just to define the default config internally (either as a raw representation of the config file contents, or as a default version of the dict structure the file is parsed to), and let an actual config file read operation overwrite those defaults.


(I did put some inline comments, but they must be read in light of this overall comment)

tests/zino_test.py
johannaengland: What should Zino do in case of an invalid `zino.toml` file? Log an error and run anyways or stop running?

runborg: as the zino.toml file will be parsed at server start time i think this i a good solution:
* In the event of non-existant zino.toml, then load all defaults and start the server. 
* if a zino.toml file exist and are parse-able parse all elements as overrides to the loaded defaults.
* on a existent but syntactli invalid zino.toml, i think the server should error out and not start with an "invalid" configuration.
also please look trough the config file for illegal options .. if this only should be printed as a warning or should be a hard error i've not concluded with at this point.. :)

johannaengland:
For documentation purposes: 
Old `zino.cf` variable names to new `zino.toml` variable names:
- `old_events` -> `archiving.old_events_dir`
- `zino_secrets` -> `authentication.file`
- `persist_file` -> `persistence.file`
- `persist_period` -> `persistence.period`
- `pollfile` -> `polling.file`
- `conf_check` -> `polling.period`

runborg:
As a reminder from our meeting last week:
If a configuration file does not exist zino should load with its internal defaults
If a configuration file is specified as a cli parameter ensure that the file exists and is a valid configuration, if the specified configuration file does not exist or includes errors zino should not start.

When loading the configuration file it should extend the internal default values definded in zino. 
Ensure that the configuration only includes known and valid configuration options, if the user have defined a key with a illegal value or if a unknown key is defined zino should not start.

stveit:
Since we now have a config we can ask for what the state  filename should be, does it make sense for  `ZinoState.load_state_from_file` to default to `STATE_FILENAME`? Is there any real usecase for the default value anyways? I imagine this function will always be called with either the value read from config or with a temp file during tests

johannaengland:
> Since we now have a config we can ask for what the state filename should be, does it make sense for `ZinoState.load_state_from_file` to default to `STATE_FILENAME`? Is there any real usecase for the default value anyways? I imagine this function will always be called with either the value read from config or with a temp file during tests

Very good point, I feel like I had a good reason for it when I was working on it to keep that default value, but I do not remember now

README.md
```diff
@@ -190,6 +190,18 @@ Zino will check `polldevs.cf` for changes on a scheduled interval while it's
 running, so any changes made while Zino is running should be picked up without
 requiring a restart of the process.
 
+### Configuring other settings
+
+Other settings can be also configured in a separate [TOML](https://toml.io/en/) file,
+which defaults to `zino.toml` in the current working directory, but a different file
+can be specified using the `--config-file` command line option.
```
lunkwill42, line 197: This is according to spec, but it's not what is currently happening in this PR:

If I create `zino.toml`, I expect Zino to read it without me specifying the `--config-file` option.  If it is missing, I expect Zino to ignore it and run with the defaults.

Currently, if I add a `./zino.toml` and run Zino without `--config-file`, Zino ignores my config.

lunkwill42 (changes requested):

src/zino/zino.py
```diff
@@ -35,7 +36,12 @@ def main():
         level=logging.INFO if not args.debug else logging.DEBUG,
         format="%(asctime)s - %(levelname)s - %(name)s (%(threadName)s) - %(message)s",
     )
-    state.state = state.ZinoState.load_state_from_file() or state.ZinoState()
+    state.config = read_configuration(args.config_file)
```
lunkwill42, line 39: Here's also a small user-friendliness annoyance: If there is a problem with the config, a full traceback is thrown in the user's face.   Exceptions here should be easy to handle and convert to a more user-friendly feedback.   Throwing a traceback at the user usually indicate some deeper unhandled problem in the code and the traceback should be sent to the developers for debugging...

lunkwill42 (changes requested):

src/zino/zino.py
```diff
@@ -41,13 +41,13 @@ def main():
         state.config = read_configuration(args.config_file or DEFAULT_CONFIG_FILE, args.polldevs)
     except OSError:
         if args.config_file:
-            print(f"No config file with the name {args.config_name} found.")
+            _log.fatal(f"No config file with the name {args.config_file} found.")
             sys.exit(1)
     except InvalidConfigurationError:
-        print(f"Configuration file with the name {args.config_name or DEFAULT_CONFIG_FILE} is invalid TOML.")
+        print(f"Configuration file with the name {args.config_file or DEFAULT_CONFIG_FILE} is invalid TOML.")
```
lunkwill42, line 47: Why still print here?

lunkwill42:
Oh well, playing fast and loose this Friday afternoon... @johannaengland left the office and I forgot there was a fixup! commit here :-D

PR #251, opened 2024-06-11
Add more codecov configuration to fix associating commits within PRs

johannaengland:
This PR does two things to hopefully get the right coverage reports for PRs from forked repos:
1. Add a checkout action to the workflow where uploading coverage happens, this is instructed in [environment specific requirements of codecov](https://docs.codecov.com/docs/environment-specific-requirements#github-actions)
2. Override the parent commit value when uploading the coverage reports by creating an artifact in the test workflow and the accessing it in the uploading workflow

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -36,12 +36,21 @@ jobs:
           files: artifacts/**/*-results.xml
 
       - name: Read PR number file
-        if: ${{ hashFiles('artifacts/pr_number/pr_number') != '' }}
+        if: ${{ hashFiles('artifacts/extra/pr_number') != '' }}
         run: |
-          pr_number=$(cat artifacts/pr_number/pr_number)
+          pr_number=$(cat artifacts/extra/pr_number)
           re='^[0-9]+$'
           if [[ $pr_number =~ $re ]] ; then
             echo "PR_NUMBER=$pr_number" >> $GITHUB_ENV
+          fi
+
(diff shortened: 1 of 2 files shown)


lunkwill42 (approved):
This looks reasonable to me, and I assume you've tested it in another repo.

Although I feel hesitant to spend even more time on this in order to get it right, I might suggest that we maybe should look into packaging this as a reusable action (if possible), so we don't need to repeat everything over and over in all our repos.

johannaengland:
> This looks reasonable to me, and I assume you've tested it in another repo.
Yes, it has been tested. It does not always show the correct number of commits, but it does show the correct coverage change due to direct and indirect changes and these can be inspected to figure out which lines have changed/missing coverage, which is what we need codecov for. 

> Although I feel hesitant to spend even more time on this in order to get it right, I might suggest that we maybe should look into packaging this as a reusable action (if possible), so we don't need to repeat everything over and over in all our repos.
I believe that that would take more time than just copy-pasting and adapting it for our other repos. Mainly because I have no experience with packaging actions and because how we do it now consists of two steps, which I don't even know if it's possible to package these into one action.

johannaengland:
A workflow file that was very helpful for me: https://github.com/edumserrano/dotnet-sdk-extensions/blob/main/.github/workflows/upload-coverage-to-codecov.yml

hmpf (approved):
My condolences. What $%^ to have in your brain...

PR #253, opened 2024-06-13
Fix serializing issues in planned maintenance

stveit:
Bugfixes for planned maintenance added in #194 (for #61)

Fixes the ol' Pydantic problem where it struggles with subclasses if you dont specifically mention the subclasses instead of just referencing the parent. Adds some default values as there were some issues reading state back in if something was set to None but didnt have default value, I'm guessing None values arent written to the file.

Also changes how the `type` field for PMs are defined, as pydantic struggled to recreate the object with the correct subclass if `Literal` wasnt being used

Does this need a towncrier file?

### DIFF
--- a/src/zino/planned_maintenance.py
+++ b/src/zino/planned_maintenance.py
@@ -40,7 +40,7 @@ class PlannedMaintenances(BaseModel):
         pm_class: type[PlannedMaintenance],
         match_type: MatchType,
         match_expression: str,
-        match_device: Optional[str],
+        match_device: Optional[str] = None,
     ) -> PlannedMaintenance:
         """Creates a planned maintenance, adds it to the planned_maintenances dict and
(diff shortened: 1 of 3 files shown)


johannaengland:
> Does this need a towncrier file?

I think it would be nice, just for the person creating the release to be able to group this and the original PM PR together

hmpf:
> > Does this need a towncrier file?
> 
> I think it would be nice, just for the person creating the release to be able to group this and the original PM PR together

If there hasn't been a release (as in: on pypi) of the pm functionality then I don't see why there needs to be something grouped together. Then again, I don't see why mentioning PR numbers in the chaneglog are necessary.

hmpf (approved):

johannaengland (approved):

Issue #247, opened 2024-06-07
Hard to find faults/typos in polldevs.cf


runborg:
When writing incorrect settings into polldevs.cf the error message presented to the user is kinda hard to read. 
there are also no information about where in the file the issue is located.
It also looks like the pydantic error field required message is truncated as then no usefull information is presented to the user

```
Traceback (most recent call last):
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/apscheduler/executors/base_py3.py", line 30, in run_coroutine_job
    retval = await job.func(*job.args, **job.kwargs)
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/zino/scheduler.py", line 70, in load_and_schedule_polldevs
    new_devices, deleted_devices = load_polldevs(polldevs_conf)
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/zino/scheduler.py", line 46, in load_polldevs
    devices = {d.name: d for d in read_polldevs(polldevs_conf)}
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/zino/scheduler.py", line 46, in <dictcomp>
    devices = {d.name: d for d in read_polldevs(polldevs_conf)}
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/zino/config/polldevs.py", line 20, in read_polldevs
    yield PollDevice(**(defaults | section))
  File "/home/runarb/zino2/zino-env/lib/python3.10/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for PollDevice
name
  Field required [type=missing, input_value={'interval': '5', 'commun...0.159', 'timeout': '11'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.5/v/missing
```

PR #254, opened 2024-06-13
Add support for subcommands in the API

### LINKED ISSUE #255: Rework the `Zino1BaseServerProtocol` command dispatcher to facilitate commands with subcommands

stveit:
Fixes #255 

Suggestion for supporting subcommands. Adds a constant tuple where commands that support subcommands can be added (for us this will contain `pm`). If a command is registered in that tuple, the dispatcher will attempt to add the first argument to the command and look for functions matching this. If main command is `pm` and subcommand is `details` then it will look for `do_pm_details`

I imagine the `Zino1ServerProtocol` subclass would register `pm` as a subcommand by overriding the tuple

This is preliminary work for implementing the planned maintenance API commands, so i dont think this needs to be part of the changelog. The commands themselves will be, however

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -34,6 +34,8 @@ def requires_authentication(func: Callable) -> Callable:
 class Zino1BaseServerProtocol(asyncio.Protocol):
     """Base implementation of the Zino 1 protocol, with a basic command dispatcher for subclasses to utilize."""
 
+    HAS_SUBCOMMANDS: tuple[str] = tuple()
+
     def __init__(
         self,
         server: Optional["ZinoServer"] = None,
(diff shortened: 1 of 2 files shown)


hmpf (commented):
Fine start, now you "just" need to update the tests :)

johannaengland (approved):

lunkwill42 (changes requested):
While it's a fine attempt, I feel like this just complicates future maintenance.  

Specifically, I'm not comfortable with having to explicitly add new methods to a an extra config tuple, located potentially far away from the implementation itself, to mark them as special.  Marking these functions using a decorator would be nicer.

However, at this point, I might suggest two different approaches:

1. Add more magic to the dispatcher code so it finds valid commands from method names, automatically understanding `do_x_y()` as an implementation of the command `X Y` while `do_x()` is a different command.

2. Or, toss the magic stuff I wrote earlier and switch to a scheme involving explicit routing of commands using decorators.  Think similarly to flask routes, e.g.
   ```python
   @route("PM")
   def do_pm(self):
       ...

   @route("PM ADD")
   def do_pm_add(self, arg1, arg2):
       ...
   ```

I would pair on something this, if desirable :)

stveit:
#247 is used instead

PR #258, opened 2024-06-17
Fix log message on initial Juniper alarm

### LINKED ISSUE #213: Yellow/Red alarm initial value "None"
On initial creation of a yellow or red alarm case, the initial state of the alarm is "None", this is not an integer value and probably should be a 0 instead. 
I do not know where in the code this value is deriving from.

```
2024-04-09 21:43:40 XXX-gw2 yellow alarms went from None to 1
2024-04-09 21:58:38 XXX-gw2 yellow alarms went from 1 to 0
2024-04-09 22:04:34 XXX-gw2 yellow alarms went from 0 to 1
```

johannaengland:
Closes #213. 

We got the log message "yellow alarms went from None to 1", because on the first Juniper alarm we create an event, which does not have an alarm count yet.

### DIFF
--- a/changelog.d/213.added.md
+++ b/changelog.d/213.added.md
@@ -1,0 +1,1 @@
+Fix log message on initial Juniper alarm
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -26,12 +26,16 @@ class JuniperAlarmTask(Task):
             }
 
         if self.device_state.alarms["yellow"] != yellow_alarm_count:
+            self.create_alarm_event(
+                color="yellow", old_alarm_count=self.device_state.alarms["yellow"], new_alarm_count=yellow_alarm_count
+            )
             self.device_state.alarms["yellow"] = yellow_alarm_count
-            self.create_alarm_event(color="yellow", alarm_count=yellow_alarm_count)
 
         if self.device_state.alarms["red"] != red_alarm_count:
+            self.create_alarm_event(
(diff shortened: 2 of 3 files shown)


hmpf:
Fun bug!

hmpf (commented):
Question: what happens if we go from N to 0 alarms?

* Is there another message?
* Is the event closed?
* What order do the two happen in, if they both happen?

Intuitively, I would assume going from N to 0 alarms for a type&device closes the alarm for that type&device.

johannaengland:
> Question: what happens if we go from N to 0 alarms?
> 
>     * Is there another message?

Yes, a log is added to the event that says exactly that.

>     * Is the event closed?

No, as in Zino1, events are never closed automatically, that has to happen manually.

hmpf:
> > ```
> > * Is the event closed?
> > ```
> 
> No, as in Zino1, events are never closed automatically, that has to happen manually.

So if it goes back from 0 to N the event is reused?

Now it makes me wonder how clients show this. 0 is a good thing and should be shown different from bigger than 0. I guess we can test when we have a way to fake an event =)

hmpf (approved):

johannaengland:
> So if it goes back from 0 to N the event is reused?

Yes, but only if the event has not been closed by then. 

> Now it makes me wonder how clients show this. 0 is a good thing and should be shown different from bigger than 0. I guess we can test when we have a way to fake an event =)

I would assume that clients show it just like any other event. But feel free to post that as an issue in Howitz

Issue #261, opened 2024-06-20
Handle errors carried via varbinds


stveit:
Some errors are carried as a varbind outside the scope of `errorStatus` and `errorIndication`.
This is what causes the issues with #212 , specifically `noSuchObject` not being handled

The different type of varbind errors are `noSuchObject`, `noSuchInstane` and `endOfMibView` as far as i can tell

PR #262, opened 2024-06-20
Add functionality for handling varbind errors in get command

### LINKED ISSUE #212: Alarm count is not int
After running the server for a while I get quite a number of devices that returns a string instead of an int alarm count.
This gives the error: 
```
Device XXX returns alarm count not of type int. Yellow alarm count: type <class 'str'>. Red alarm count: type <class 'str'>.
```
the error does not give any indications of whats inside the value retrieved. if this is a unsupported feature on the platform or if there are other issues? 

Looking into this it looks like the same routers reporting this error and the "router count" is stable... 
```
(zino-env) --> $ cat log  | grep ERROR | grep "Yellow alarm count:" | cut -d\  -f 9- | sort | uniq -c | sort -h | wc -l
48
```

48 out of our routers returns this, and i'm able to produce a list of affected routers if there is interest in this for troubleshooting.

stveit:
Work towards fixing #261 and #212 

Adds function that takes an `ObjectType` and raises a relevant exception if it contains an error instead of a real value. I refer to this type of error as varbind errors since they are not defined via `errorStatus` or `errorIndication`, but are defined as values inside the response varbind

This PR makes use of the new function in the `get` command, but should be used for all the different commands. That can be done in other PRs once this one is completed.

To fully fix #212, some changes have to be made to the `juniperalarmtask`, including reverting some changes made in #231.
Handling of the new errors should also be added. `NoSuchObjectError` is definitely relevant, but maybe `NoSuchInstanceError` is too? Im not too sure about the difference about these two.. I'll leave @johannaengland to complete this task

Also sneakily fixes `ErrorIndication` not inheriting correctly (an SNMP related exception should probably inherit from the exception specifically made to be a base exception for everything SNMP related...)

Also2, there's quite a few SNMP specific exception classes now, thoughts on putting them in their own file?

### DIFF
--- a/changelog.d/261.fixed.md
+++ b/changelog.d/261.fixed.md
@@ -1,0 +1,1 @@
+Add handling of varbind errors for snmp get actions.
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -142,7 +142,9 @@ class SNMP:
         except PysnmpMibNotFoundError as error:
             raise MibNotFoundError(error)
         self._raise_errors(error_indication, error_status, error_index, query)
-        return self._object_type_to_mib_object(var_binds[0])
+        result = var_binds[0]
+        self._raise_varbind_errors(result)
+        return self._object_type_to_mib_object(result)
 
     def _raise_errors(
         self,
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -75,7 +75,7 @@ class SnmpError(Exception):
     """Base class for SNMP, MIB and OID specific errors"""
 
 
-class ErrorIndication(Exception):
+class ErrorIndication(SnmpError):
     """Class for SNMP errors that occur locally,
     as opposed to being reported from a different SNMP entity.
     """
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -21,7 +21,7 @@ from pysnmp.hlapi.asyncio import (
     nextCmd,
 )
 from pysnmp.proto import errind
-from pysnmp.proto.rfc1905 import errorStatus
+from pysnmp.proto.rfc1905 import EndOfMibView, NoSuchInstance, NoSuchObject, errorStatus
 from pysnmp.smi import builder, view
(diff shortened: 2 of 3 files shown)


lunkwill42:
> Work towards fixing #261 and #212
> 
> Adds function that takes an `ObjectType` and raises a relevant exception if it contains an error instead of a real value. I refer to this type of error as varbind errors since they are not defined via `errorStatus` or `errorIndication`, but are defined as values inside the response varbind
> 
> This PR makes use of the new function in the `get` command, but should be used for all the different commands. That can be done in other PRs once this one is completed.
> 
> To fully fix #212, some changes have to be made to the `juniperalarmtask`, including reverting some changes made in #231. Handling of the new errors should also be added. `NoSuchObjectError` is definitely relevant, but maybe `NoSuchInstanceError` is too? Im not too sure about the difference about these two.. I'll leave @johannaengland to complete this task
> 
> Also sneakily fixes `ErrorIndication` not inheriting correctly (an SNMP related exception should probably inherit from the exception specifically made to be a base exception for everything SNMP related...)
> 
> Also2, there's quite a few SNMP specific exception classes now, thoughts on putting them in their own file?

lunkwill42 (approved):
> This PR makes use of the new function in the `get` command, but should be used for all the different commands. That can be done in other PRs once this one is completed.

Not necessarily. In cases where the client code expects to fetch multiple values using a single operation, the client may be fine with *some* of the return values being errors.  The client might want to keep the successfully retrieved values and make decisions about changing its queries for failed values.

This could also be a flag to a method that retrieves multiple values.  The default could be to raise exceptions, but if the calling code wants to handle error values in custom ways, it could set a flag, asking the library to stay out of its business.

> Also sneakily fixes `ErrorIndication` not inheriting correctly (an SNMP related exception should probably inherit from the exception specifically made to be a base exception for everything SNMP related...)

:+1: 

> Also2, there's quite a few SNMP specific exception classes now, thoughts on putting them in their own file?

Sure, I'd support that - but the likely cleanup here would be to split `zino/snmp.py` into `zino/snmp/__init__.py` and `zino/snmp/errors.py`.

PR #265, opened 2024-06-21
Add neigh_rdns to BFDEvent

### LINKED ISSUE #199: BFD: missing attribute neigh_rdns
The attribute `neigh_rdns` should include "reverse dns" information of the BFD neighbor. (reverse dns of the address in the bfdaddr field i think)

the zino1 server is not consistent on this field on all types containing ipv4/ipv6 neighbor info, but this could certanly also be done on eg. BGP alarms.

This field is not critical for eg. curitz to function at this state

stveit:
Fixes #199

### DIFF
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -350,6 +350,7 @@ class BFDEvent(Event):
     bfdix: Optional[int] = None
     bfddiscr: Optional[int] = None
     bfdaddr: Optional[IPAddress] = None
+    neigh_rdns: Optional[str] = None
 
     @property
     def subindex(self) -> SubIndex:
--- a/src/zino/utils.py
+++ b/src/zino/utils.py
@@ -1,4 +1,6 @@
+import socket
 from ipaddress import ip_address
+from typing import Optional
 
 from pyasn1.type.univ import OctetString
 
@@ -25,3 +27,10 @@ def _parse_hexa_string_ip(ip: str) -> IPAddress:
 def _parse_colon_separated_ip(ip: str) -> IPAddress:
     """Parses IP addresses formatted with a colon symbol separating every octet, e.g. 7F:00:00:01"""
     return ip_address(bytes(OctetString(hexValue=ip.replace(":", ""))))
+
+
+def reverse_dns(ip: str) -> Optional[str]:
+    try:
+        return socket.gethostbyaddr(str(ip))[0]
+    except socket.herror:
+        return None
--- a/src/zino/utils.py
+++ b/src/zino/utils.py
@@ -1,7 +1,8 @@
-import socket
+import asyncio
 from ipaddress import ip_address
 from typing import Optional
 
+import aiodns
 from pyasn1.type.univ import OctetString
 
 from zino.statemodels import IPAddress
@@ -29,8 +30,12 @@ def _parse_colon_separated_ip(ip: str) -> IPAddress:
     return ip_address(bytes(OctetString(hexValue=ip.replace(":", ""))))
 
 
-def reverse_dns(ip: str) -> Optional[str]:
+async def reverse_dns(ip: str) -> Optional[str]:
+    """Returns hostname for given IP address or None if reverse DNS lookup fails"""
+    loop = asyncio.get_event_loop()
+    resolver = aiodns.DNSResolver(loop=loop)
     try:
-        return socket.gethostbyaddr(str(ip))[0]
-    except socket.herror:
+        response = await resolver.gethostbyaddr(ip)
+        return response.name
+    except aiodns.error.DNSError:
         return None
--- a/changelog.d/199.added.md
+++ b/changelog.d/199.added.md
@@ -1,0 +1,1 @@
+Add support for neigh_rdns in BFD events.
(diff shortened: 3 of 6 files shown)


src/zino/utils.py
```diff
@@ -25,3 +27,10 @@ def _parse_hexa_string_ip(ip: str) -> IPAddress:
 def _parse_colon_separated_ip(ip: str) -> IPAddress:
     """Parses IP addresses formatted with a colon symbol separating every octet, e.g. 7F:00:00:01"""
     return ip_address(bytes(OctetString(hexValue=ip.replace(":", ""))))
+
+
+def reverse_dns(ip: str) -> Optional[str]:
+    try:
+        return socket.gethostbyaddr(str(ip))[0]
+    except socket.herror:
+        return None
```
lunkwill42, line 36: This is a blocking socket operation and does not belong in an async event loop at all.

If we need to do async DNS requests, we need to look into other libraries, maybe something like `aiodnsresolver`.

lunkwill42 (changes requested):
The blocking socket is a show-stopper for this.

Otherwise, I'd ask about the original issue as well. I could not find `neigh_rdns` in the original Zino 1 codebase, but rather it seems the event attribute is named `Neigh-rDNS` (verbatim).  Does this matter, @runborg ?

src/zino/utils.py
lunkwill42: The most popularly used library I've been able to find in a few short minutes of research is https://github.com/saghul/aiodns

stveit: derp, made some changes now

src/zino/statemodels.py
```diff
@@ -350,6 +350,7 @@ class BFDEvent(Event):
     bfdix: Optional[int] = None
     bfddiscr: Optional[int] = None
     bfdaddr: Optional[IPAddress] = None
+    neigh_rdns: Optional[str] = None
```
johannaengland, line 353: According to @runborg  the name needs to be `Neigh-rDNS` ref: https://github.com/Uninett/zino/issues/199#issuecomment-2186359376
```suggestion
    Neigh-rDNS: Optional[str] = None
```

johannaengland (dismissed):

src/zino/statemodels.py
lunkwill42: Indeed, except this is an invalid name in Python, so this doesn't work.  We may have to look into Pydantic field aliases.  One thing is supporting them on deserialization, but I can't recall what we do for serializing to the API, I'll have to check

runborg:
Looking at this there could potentially be multiple dns resolution requests in short time, i think this could be speeded up by using some kind of caching. eg. `lru_cache` or `ttl_cache`

stveit:
> Looking at this there could potentially be multiple dns resolution requests in short time, i think this could be speeded up by using some kind of caching. eg. `lru_cache` or `ttl_cache`

Caching can be added. Should it then have a time-to-live, a certain duration where each result will be deleted after existing for a certain period of time?

stveit:
> Looking at this there could potentially be multiple dns resolution requests in short time, i think this could be speeded up by using some kind of caching. eg. `lru_cache` or `ttl_cache`

Caching can be added. How long do you imagine the TTL should be? and any ideas for maxsize to use?

src/zino/statemodels.py
stveit: This is the case for several values (`alarm_count` and `alarm_type` in AlarmEvent for instance). A simple solution would be to "zinofy" the attribute name in `model_dump_simple_attrs` by replacing `_` with `-`

lunkwill42:
> Caching can be added. How long do you imagine the TTL should be? and any ideas for maxsize to use?

I can't see that Zino 1 does any caching here, so I'm not sure we should bother with this for Zino 2.0.  Post-zino-2 would be fine.

src/zino/statemodels.py
lunkwill42: True, `model_dump_simple_attrs` exists specifically for the legacy protocol - so it should indeed replace `_` with `-` in the output (are there any attrs in the protocol that use an underscore?)

However, that still isn't enough if the *casing* is also important for the protocol clients, which makes me think a Field alias might be the way to go.  `model_dump_simple_attrs` could look up field aliases and use them if they exist, in addition to replacing underscores.

lunkwill42: On the other hand, question for @runborg : I cannot see that `Neigh-rDNS` is referred to anywhere else in the Zino Tcl code, not even in the Ritz tree (even when searching case insensitively).  Which again leads me to wonder: Does the casing really matter to anybody?

stveit:
> > Caching can be added. How long do you imagine the TTL should be? and any ideas for maxsize to use?
> 
> I can't see that Zino 1 does any caching here, so I'm not sure we should bother with this for Zino 2.0. Post-zino-2 would be fine.

:+1:

src/zino/statemodels.py
stveit: Casing is a good point, if that turns out to matter then something per-field is prob required.

I'd say this outside the scope of this PR anyways, would be weird to introduce special handling of event attributes for neigh_rdns only

lunkwill42 (approved):
I made a separate issue for the API output stuff discussed in the comments :)

Issue #243, opened 2024-06-03
Have an easy tool to fake an event, for debugging/testing clients


hmpf:
The tool needs to be able to

1. create a new event (see if it turns up in the correct location)
2. close an existing event (admstate closed)
3. remove an existing (closed) event on the server (to see if it disappears in clients)

Changing admstate can be tested via existing clients so we do not need support for that.

Workaround: in-depth docs for how to do this via telnet though point 3 is probably not possible via telnet.

johannaengland:
We decided to make this into a new API command and I think it is actually best to split it into three parts:
* [ ] Implement an API command that creates fake events
* [ ] Implement an API command that sets the state of existing events by the given id
* [ ] Implement an API command that deletes an event by the given id

runborg:
as for point 2) the `SETSTATE` [api command](https://github.com/Uninett/zino/blob/58bf6f8a4296c287af2c2ca9bdfe9c21bcd2ae8d/src/zino/api/legacy.py#L329) are able to set the state of an event, if you choose the state `CLOSED` it will be closed, if this command is usable use that instead reimplement it in #272

PR #273, opened 2024-06-24
Add API command to delete event by given id

### LINKED ISSUE #270: Implement an API command that deletes an event by the given id

johannaengland:
Closes #270. Last part to #243.


src/zino/api/legacy.py
```diff
@@ -619,3 +620,21 @@ async def do_raiseerror(self):
...
+        the frontend.
+        """
+        events = self._state.events
+
+        index = EventIndex(event.router, event.subindex, type(event))
+
+        if events._events_by_index.get(index) and event.id == events._events_by_index[index].id:
+            del events._events_by_index[index]
+        if events._closed_events_by_index.get(index) and event.id == events._closed_events_by_index[index].id:
+            del events._closed_events_by_index[index]
```
stveit, line 637: If the point of the first check is to see if the dict has a value for a certain index then I think this is a better way of writing it:
```suggestion
        if index in events._events_by_index and event.id == events._events_by_index[index].id:
            del events._events_by_index[index]
        if index in events._closed_events_by_index and event.id == events._closed_events_by_index[index].id:
            del events._closed_events_by_index[index]
```

tests/api/legacy_test.py
```diff
@@ -892,6 +894,55 @@ async def test_when_authenticated_should_list_all_pm_ids(self, authenticated_pro
...
+        event = ReachabilityEvent(router=DEVICE_NAME, state=EventState.CLOSED)
+        protocol._state.events.commit(event=event)
+
+        command = f"DELETEEVENT {event.id}"
+        await protocol.message_received(command)
+
+        assert fake_transport.write.called
+        response = fake_transport.write.call_args[0][0].decode("utf-8")
+        assert response.startswith("200 ")
+        assert response.startswith("200 ")
```
stveit, line 938: dupe

src/zino/api/legacy.py
johannaengland: Thank you, very good point!

PR #274, opened 2024-06-24
Rewrite API command dispatcher to match commands by patterns

### LINKED ISSUE #255: Rework the `Zino1BaseServerProtocol` command dispatcher to facilitate commands with subcommands

lunkwill42:
An alternative to #254.

This changes the API command dispatcher framework more fundamentally to pre-build maps of API command patterns to match incoming messages against, in order to support arbitrarily long "sub-command" names.

With these changes, a `PM` command with e.g. a `PM FOO` sub-command can be implemented by adding two methods, following the existing pattern, and with no need to "pre-register" as a sub-command implementation:

```python
def do_pm(self):
   ...

def do_pm_foo(self, arg1, arg2, ...):
  ...
```

@stveit: You weren't available for pair programming, so I threw this together at my own initiative.

Fixes #255

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -64,6 +64,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         self._multiline_future: asyncio.Future = None
         self._multiline_buffer: List[str] = []
         self._authentication_challenge: Optional[str] = None
+        self._responders = self._get_all_responders()
 
         self._state = state if state is not None else ZinoState()
         self._secrets_file = secrets_file
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -125,27 +125,26 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
 
         if not message:
             return
-        args = message.split(" ")
-        return self._dispatch_command(*args)
+        return self._dispatch_command(message)
 
-    def _dispatch_command(self, command, *args):
-        responder = self._get_responder(command)
+    def _dispatch_command(self, message: str):
+        responder, args = self._get_responder(message)
         if not responder:
-            return self._respond_error(f'unknown command: "{command}"')
+            return self._respond_error(f'unknown command: "{message}"')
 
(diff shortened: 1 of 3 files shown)


src/zino/api/legacy.py
```diff
@@ -235,13 +235,15 @@ async def do_quit(self):
...
-            }
-
-        commands = " ".join(sorted(responders))
+        """Lists all available top-level API commands"""
+        eligible = (
+            responder.name
+            for responder in self._responders.values()
+            if " " not in responder.name
+            and (self.is_authenticated or not getattr(responder.function, "requires_authentication", False))
+        )
```
johannaengland, line 244: I would prefer this ultra long condition to be split into three parts with descriptive names, because it took me 5 minutes to wrap my head around this

johannaengland (changes requested):
I would like to have at least one more test for `_get_responder`, one that shows, that it actually returns the longer command in case there is a subcommand

src/zino/api/legacy.py
lunkwill42: If you consider three expressions in boolean logic to be *ultra* long, you will have issues next time you look at legacy code :grin:

johannaengland: Maybe I was exaggerating a little :grin:
But the nice thing about this re-implementation is that it doesn't have to be like legacy code :smile:

lunkwill42:
> And we should document in the [Developing Zino](https://github.com/Uninett/zino?tab=readme-ov-file#developing-zino) part of the docs that sub commands are responsible to implement a sub HELP command if they want one

Actually, I'm not sure I care to spend time on documenting this as long as our long-term goal is to replace the legacy API with more modern constructs.  The only known sub-command implementation thus far is the `PM` command, and my hope is we never will expand beyond this ;)

src/zino/api/legacy.py
lunkwill42: Clarifications and test addition ready for review :)

johannaengland (approved):
Beautiful, thank you very much

stveit (approved):

PR #275, opened 2024-06-25
Catch new SNMP errors in juniper alarm task

### LINKED ISSUE #212: Alarm count is not int

johannaengland:
## Scope and purpose

Fixes #212.

This amends the changes made in #231 according to the new SNMP errors introduced/handled in #262.
This should finally lead to error being correctly handled and only things being logged if the device has an alarm count of an invalid type (but not none).

## Contributor Checklist

### DIFF
--- a/src/zino/tasks/juniperalarmtask.py
+++ b/src/zino/tasks/juniperalarmtask.py
@@ -1,6 +1,6 @@
 import logging
 
-from zino.snmp import NoSuchNameError
+from zino.snmp import NoSuchNameError, VarBindError
 from zino.statemodels import AlarmEvent, AlarmType
 from zino.tasks.task import Task
 
@@ -16,7 +16,7 @@ class JuniperAlarmTask(Task):
 
         try:
             yellow_alarm_count, red_alarm_count = await self._get_juniper_alarms()
-        except (TypeError, NoSuchNameError):
+        except (NoSuchNameError, TypeError, VarBindError):
             return
 
         if not self.device_state.alarms:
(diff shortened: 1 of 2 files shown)


src/zino/tasks/juniperalarmtask.py
```diff
@@ -16,7 +21,7 @@ async def run(self):
 
         try:
             yellow_alarm_count, red_alarm_count = await self._get_juniper_alarms()
-        except (TypeError, NoSuchNameError):
+        except (EndOfMibViewError, NoSuchInstanceError, NoSuchNameError, NoSuchObjectError, TypeError):
```
lunkwill42, line 24: Might as well utilize the known exception class hierarchies here...

```suggestion
        except (NoSuchNameError, TypeError, VarBindError):
```

src/zino/tasks/juniperalarmtask.py
```diff
@@ -1,6 +1,11 @@
 import logging
 
-from zino.snmp import NoSuchNameError
+from zino.snmp import (
+    EndOfMibViewError,
+    NoSuchInstanceError,
+    NoSuchNameError,
+    NoSuchObjectError,
+)
```
lunkwill42, line 8:
```suggestion
from zino.snmp import NoSuchInstanceError, VarBindError
```

lunkwill42 (changes requested):

src/zino/tasks/juniperalarmtask.py
johannaengland: Good call, didn't think of that!

lunkwill42 (approved):
:+1:

PR #277, opened 2024-06-26
Index closed events

lunkwill42:
## Scope and purpose

The original Zino code will sometimes alter already closed events.  Apparently, the reasoning goes that operators sometimes prematurely close events for flapping ports (which Håvard E argues that we should maybe disallow in the future).  The original Zino therefore takes care to update both a closed flapping event and the currently open flapping event whenever the link trap code handles a flapping state.

A limitation I immediately see is that multiple closed events could exist for a flapping state (until they are expired/scavenged), but we can only index one of them.  As with the original code, there is a bias towards only indexing the latest event.

As a side note: Zino 1 used the terms "pre-closed" and "closed" for events.  The former refers to events that have been given a state of CLOSED by an operator.  The latter refers to events that have been evicted completely from the running state.

### This pull request

* Adds a `get_closed_event()` method to the `Events` class.
* Adds internal indexing of closed events to the `Events` class.

## Contributor Checklist

  * Not sure I need this for this feature, which is strictly internal.  No effect for end users (yet).

### DIFF
--- a/changelog.d/+closed-events-index.changed.md
+++ b/changelog.d/+closed-events-index.changed.md
@@ -1,0 +1,1 @@
+Index recently closed events to facilitate updating of prematurely closed flapping events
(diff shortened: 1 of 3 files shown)


johannaengland (approved):
:+1:

Issue #288, opened 2024-07-04
Implement handling of Juniper BGP traps (`jnxBgpM2BackwardTransition` and `jnxBgpM2Established` from `BGP4-V2-MIB-JUNIPER`)


lunkwill42:
Zino 1 appears to ignore traps from the regular IETF `BGP4-MIB`, and only handles traps from the Juniper-specific BGP MIB `BGP4-V2-MIB-JUNIPER`.

Mainly two traps are handled:
- `jnxBgpM2BackwardTransition` - indicates that a BGP peer was lost.
- `jnxBgpM2Established` - indicates that a BGP peer came back up.

lunkwill42:
Closed by #291

PR #290, opened 2024-07-04
Resolve incoming trap varbinds properly

lunkwill42:
## Scope and purpose

The varbinds of an incoming trap message were incorrectly resolved (most likely due to a misunderstanding of PySNMP complexity).

In fact, the original resolver only resolved the name of the incoming variable, but the subtype of the value also needs to be resolved according to the MIB definition of the incoming variable.

This new resolver ensures the value is also resolved to the correct type. An example: The incoming raw value may be an ASN1 OctetString, but seen in light of the resolved MIB object name, it might actually be resolved as a InetAddress textual convention type instead (which would heavily influence how the incoming value is parsed to a Python value).

This issue was discovered during implementation of #288 as incoming BGP operational state values were not translated to names, but stayed as opaque integers

### This pull request
* changes the internal implementation of the trap varbind resolver, but no external usage is affected (except for improved ability to resolve incoming value types according to MIBs)
## Contributor Checklist

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       # Git Checkout
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0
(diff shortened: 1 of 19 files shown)


lunkwill42:
No comments received in an IRL code review within our team, so I'm merging this now.

PR #300, opened 2024-07-05
Add POLLINTF API command

### LINKED ISSUE #218: Implement the `POLLINTF` API command
Input: router name and interface index

Should asynchronously call `poll_single_interface` for that router and interface, does not wait for the response, simply fire and forget and returns `200 OK` in that case. Returns `500 Router %s unknown` in case of the router not being defined in the pollfile.

johannaengland:
## Scope and purpose

Fixes #218.

### This pull request
* changes the API

## Contributor Checklist

  * well, tests are documentation...

### DIFF
--- a/changelog.d/300.added.md
+++ b/changelog.d/300.added.md
@@ -1,0 +1,1 @@
+Add `POLLINTF` API command
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -9,7 +9,7 @@ import inspect
 import logging
 import re
 import textwrap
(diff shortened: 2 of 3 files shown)


lunkwill42 (approved):
I've taken the liberty to clean this up a little in your absence.

My changes made me realize that we should really consider adding a polldevs dict to the API server instance as well.  The original reached into the global state, which isn't necessary, since the API server instance has a reference to the state.  However, it still reaches out to the global `polldevs` dictionary, which I would prefer it didn't.

PR #305, opened 2024-07-08
Implement a `BFDTrapObserver`

### LINKED ISSUE #292: Implement handling of BFD traps (`BFD-STD-MIB::bfdSessUp` and `BFD-STD-MIB::bfdSessDown`)
Implement support for the two `BFD-STD-MIB` notifications/traps:

```asn1
bfdSessUp NOTIFICATION-TYPE
    OBJECTS {
        bfdSessDiag, -- low range value
        bfdSessDiag  -- high range value
    }
    STATUS     current
    DESCRIPTION
        "This notification is generated when the
         bfdSessState object for one or more contiguous
         entries in bfdSessTable are about to enter the up(4)
         state from some other state.  The included values of
         bfdSessDiag MUST both be set equal to this
         new state (i.e., up(4)).  The two instances of
         bfdSessDiag in this notification indicate the range
         of indexes that are affected.  Note that all the indexes
         of the two ends of the range can be derived from the
         instance identifiers of these two objects.  For the
         cases where a contiguous range of sessions
         have transitioned into the up(4) state at roughly
         the same time, the device SHOULD issue a single
         notification for each range of contiguous indexes in
         an effort to minimize the emission of a large number
         of notifications.  If a notification has to be
         issued for just a single bfdSessEntry, then
         the instance identifier (and values) of the two
         bfdSessDiag objects MUST be identical."
    ::= { bfdNotifications 1 }

bfdSessDown NOTIFICATION-TYPE
    OBJECTS {
        bfdSessDiag, -- low range value
        bfdSessDiag  -- high range value
    }
    STATUS     current
    DESCRIPTION
        "This notification is generated when the
         bfdSessState object for one or more contiguous
         entries in bfdSessTable are about to enter the down(2)
         or adminDown(1) states from some other state.  The included
         values of bfdSessDiag MUST both be set equal to this new
         state (i.e., down(2) or adminDown(1)).  The two instances
         of bfdSessDiag in this notification indicate the range
         of indexes that are affected.  Note that all the indexes
         of the two ends of the range can be derived from the
         instance identifiers of these two objects.  For
         cases where a contiguous range of sessions
         have transitioned into the down(2) or adminDown(1) states
         at roughly the same time, the device SHOULD issue a single
         notification for each range of contiguous indexes in
         an effort to minimize the emission of a large number
         of notifications.  If a notification has to be
         issued for just a single bfdSessEntry, then
         the instance identifier (and values) of the two
         bfdSessDiag objects MUST be identical."
    ::= { bfdNotifications 2 }
```

This means that the handler will mostly have to defer to the existing `BFDTask` code to poll the updated state of individual sessions when these traps are received.

lunkwill42:
## Scope and purpose

Fixes #292.  Dependent on #303 to be merged first.

`bfdSessUp` and `bfdSessDown` traps only tell which BFD sessions have changed state and why, but does not actually include the new state value for the session.  This means most of this new observer's work is to just offload everything to `BFDTask` to poll the individual session(s) that have changed state and ensure state and events are updated accordingly.

### This pull request

* Adds a new BFD Trap observer to the default Zino trap listener setup.
* Adds the optional ability for the existing `BFDTask` to fetch and update single BFD sessions, used by the trap observer.

## Contributor Checklist

### DIFF
--- a/changelog.d/305.added.md
+++ b/changelog.d/305.added.md
@@ -1,0 +1,1 @@
+Act to update BFD session information on incoming BFD session traps
--- a/src/zino/tasks/bfdtask.py
+++ b/src/zino/tasks/bfdtask.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict
+from typing import Dict, Sequence
 
 from zino.oid import OID
 from zino.scheduler import get_scheduler
@@ -35,12 +35,15 @@ class BFDTask(Task):
         super().__init__(*args, **kwargs)
         self._scheduler = get_scheduler()
 
-    async def run(self):
+    async def run(self, session_index: int = None):
+        """Polls data for all BFD sessions.  If session_index is provided, only that session is polled."""
+        msg = f" for session index {session_index}" if session_index is not None else ""
+        _log.debug("%s: polling BFD data%s", self.device.name, msg)
         if self.device_state.is_juniper:
-            polled_state = await self._poll_juniper()
+            polled_state = await self._poll_juniper(session_index)
             await self._update_state_for_all_ports_juniper(polled_state)
         elif self.device_state.is_cisco:
-            polled_state = await self._poll_cisco()
+            polled_state = await self._poll_cisco(session_index)
             await self._update_state_for_all_ports_cisco(polled_state)
 
     async def _update_state_for_all_ports_juniper(self, polled_state: DescrBFDStates):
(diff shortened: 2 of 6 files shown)


johannaengland (approved):
But I would like to have an updated news fragment - act is a strange choice here. IMO it makes more sense to simply remove `Act to` from it.

PR #316, opened 2024-07-17
Pin `snmpsim` to an older version

lunkwill42:
## Scope and purpose

A new version of snmpsim was released about 6 hours ago (the first new version in 5 years), and it seems to have changed to be incompatible with how our test suite uses it.

In particular, our test suite expects `snmpsimd.py` to be available on the search PATH, but the executable's name may have changed, and I don't have the luxury of time to figure out how to adapt our test suite

I assume there may be many more changes, as the original maintainer died a while back, and someone else has taken over, which is why I think it's best to pin to an older version for now.

### This pull request
* pins the `snmpsim` dependency to a version less than `1.0`

## Contributor Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ dev = [
     "pytest-timeout",
     "retry",
     "ruff",
-    "snmpsim",
+    "snmpsim<1.0",
     "towncrier",
     "tox<4",
     "twine",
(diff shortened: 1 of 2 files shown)


lunkwill42 (commented):
Test suite is green now, merging.

PR #327, opened 2024-07-26
Only load and parse pollfile if it has changed since last loading

### LINKED ISSUE #282: Only load and parse `polldevs.cf` if its `mtime` value has changed
This is an optimization employed by Zino 1, so Zino 2 should do it as well:

The `polldevs.cf` file is read and parsed on a set interval.  Differences between `polldevs.cf` and the running configuration are detected and executed internally.

However, repeatedly parsing the full `polldevs.cf` is a waste of time if it rarely changes (time that could be spent polling devices, rather than blocking the event loop).  The reloading routine should keep track of the modified-time (`mtime`) of the `polldevs.cf` file, and only initiate a reload if it detects that the file has been modified since it was last read.

johannaengland:
## Scope and purpose

Fixes #282. 

Add a description of what this PR does and why it is needed. If a linked ticket(s) fully
cover it, you can omit this.

## Contributor Checklist


src/zino/scheduler.py
```diff
@@ -40,14 +42,27 @@ def get_scheduler() -> AsyncIOScheduler:
...
     """
+    if last_run_time:
+        try:
+            st_mtime = pathlib.Path(polldevs_conf).stat().st_mtime
+            changed_time = datetime.fromtimestamp(st_mtime, tz=timezone.utc)
+        except OSError as e:  # noqa
+            _log.error(e)
+            return set(), set()
+
+        if changed_time <= last_run_time:
```
runborg, line 60: If the user replaces the polldevs file with a old copy using eg. the `mv` command, the mtime (modified time) parameter is not changed and will be a timestamp in the past. zino should still consider this as a change and load the older polldevs file.  the solution here is to only skip reading if the new mtime is exactly the same as the previous  mtime

runborg (dismissed):

src/zino/scheduler.py
runborg: i think the correct thing here is to only take mtime into account, and not the timestamp when the config was read. The test will then be to check if mtime is changed and not if its newer than the timestamp when the last read was done.

johannaengland: I had not considered that! Very good point, will change that accordingly!

johannaengland:
I have rebased this instead of adding more commit because I have rethought how to access the saved attribute of when the poll file was last modified.

src/zino/state.py
```diff
@@ -25,6 +26,9 @@
 
 config: Configuration = Configuration()
 
+# Last time the pollfile was modified
+pollfile_mtime: Optional[datetime] = None
```
lunkwill42, line 30: This is purely an internal value, so I'm not sure why we bother converting `mtime` from a unix timestamp to a `datetime` object at all.  We're really just verifying whether the value changed at all, so it doesn't even matter what the value represents.

lunkwill42 (changes requested):
Nice! Still, let me throw some nit-picks at you :)

src/zino/state.py
johannaengland: I found it easier having it as a datetime object for debugging, but if you think we should keep it as a unix timestamp, we can do that

src/zino/scheduler.py
```diff
@@ -41,10 +42,21 @@ def get_scheduler() -> AsyncIOScheduler:
...
-    """Loads pollfile into process state.
+    """Loads pollfile into process state if it was changed since the last time it
+    was loaded
 
     :returns: A tuple of (new_devices, deleted_devices, changed_devices and a dictionary of default settings)
     """
+    try:
+        st_mtime = pathlib.Path(polldevs_conf).stat().st_mtime
+        modified_time = datetime.fromtimestamp(st_mtime, tz=timezone.utc)
+    except OSError as e:  # noqa
```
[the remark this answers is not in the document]
johannaengland: I think that is leftover from a previous version, will remove it

lunkwill42 (approved):
Might still consider factoring out the timestamp checking to a separate function, but this will definitely do :)

PR #333, opened 2024-07-31
Reschedule changed devices

### LINKED ISSUE #330: Updating default interval in polldevs does not re-schedule tasks
after a change of `default interval` in polldevs the jobs are not re-scheduled by the new interval specified.
(This might also be the case for `interval` statement on a device, i have not tested that yet)
the new interval will only be used after a complete reinitialization of zino

this can be observed by looking at the `Running job "......` messages and noting the `interval[x:xx:xx]` stamps
Then change default interval in polldevs and take a new look in the log.

on a reread of config, are collection jobs for the affected nodes re-initiated correctly?

johannaengland:
## Scope and purpose

Fixes #330.

Makes `load_polldevs` detect changed devices and then deletes scheduled jobs for these and adds new jobs with the new configuration values.

## Contributor Checklist

### DIFF
--- a/src/zino/scheduler.py
+++ b/src/zino/scheduler.py
@@ -40,30 +40,35 @@ def get_scheduler() -> AsyncIOScheduler:
 
 
 @log_time_spent()
-def load_polldevs(polldevs_conf: str) -> Tuple[Set, Set]:
+def load_polldevs(polldevs_conf: str) -> Tuple[Set, Set, Set]:
     """Loads pollfile into process state.
 
-    :returns: A tuple of (new_devices, deleted_devices)
+    :returns: A tuple of (new_devices, deleted_devices, changed_devices)
     """
     try:
         devices = {d.name: d for d in read_polldevs(polldevs_conf)}
     except InvalidConfiguration as error:
         _log.error(error)
-        return set(), set()
+        return set(), set(), set()
 
     new_devices = set(devices) - set(state.polldevs)
     deleted_devices = set(state.polldevs) - set(devices)
+    overlap = set(state.polldevs) - deleted_devices
+    changed_devices = {device for device in overlap if devices[device] != state.polldevs[device]}
+
     if new_devices:
         _log.info("loaded new devices: %r", new_devices)
         init_state_for_devices((devices[d] for d in new_devices))
     if deleted_devices:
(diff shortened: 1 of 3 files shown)


src/zino/scheduler.py
```diff
@@ -74,9 +79,9 @@ def init_state_for_devices(devices: Sequence[PollDevice]):
 
 
 async def load_and_schedule_polldevs(polldevs_conf: str):
-    new_devices, deleted_devices = load_polldevs(polldevs_conf)
-    schedule_new_devices(new_devices)
-    deschedule_deleted_devices(deleted_devices)
+    new_devices, deleted_devices, changed_devices = load_polldevs(polldevs_conf)
+    deschedule_deleted_devices(deleted_devices | changed_devices)
+    schedule_new_devices(new_devices | changed_devices)
```
stveit, line 84: `schedule_new_devices` will make a log message saying `_log.debug("Scheduling %s new devices", len(devices))`, which will be a bit misleading when some of the "new devices" are existing devices that are being re-scheduled in order to update them. Could these two be separated somehow, that its clear that some are new devices and some are existing ones that are updated?

johannaengland: Would it be enough in your opinion to change the log message to `Scheduling %s new/changed devices`? Because in `load_polldevs` we are logging which devices are new and which are changed (`_log.info("loaded new devices: %r", new_devices)`), so we will have that information just above in the logs.

johannaengland: Or I could write a separate function that does actually change the existing jobs instead of deleting the existing jobs and creating new ones.

stveit: `new/changed` is probably fine in that case

src/zino/scheduler.py
```diff
@@ -74,17 +79,17 @@ def init_state_for_devices(devices: Sequence[PollDevice]):
...
```
stveit, line 84: What do you think about renaming these functions now that they are used for more than just new and deleted devices?
Renaming to `deschedule_devices` and `schedule_devices` might be better now. Then we can change the log message again to just be `_log.debug("Scheduling %s devices", len(devices))`

johannaengland: I think that is a very good idea

stveit (approved):

PR #339, opened 2024-08-02
Use default interval from pollfile to stagger new jobs

### LINKED ISSUE #337: Scheduling interval of new jobs fixed to 300sec.
When zino initialises or jobs are reinitialised all jobs are started in a sequence delayed by the number of nodes divided by the internal default reschedule interval.  As for now this is hardcoded to 5minutes / 300 sec.

Increasing the default interval in polldevs have no effect on this scheduling, so on a overloaded server increasing the defined retry interval (`default interval`) to lower the load results in the server getting overloaded in the start of the interval and be idle when all scheduled jobs are finished running. The reason for this is that all jobs are initiated in the first 300 sec of the interval.

This should be changed to follow the polldev defined `default interval` if the user have defined this.

https://github.com/Uninett/zino/blob/41a87dcd86fbfdde26c063b267001f11a2ef5543/src/zino/scheduler.py#L91-L94

johannaengland:
## Scope and purpose

Fixes #337.

Changed the way that `read_polldevs` work to make it easier to return the `PollDevices` and the default values. I was considering splitting the function in two: One to read the defaults and one to read the rest - the poll devices, but decided for this method, since it is simpler changes.

## Contributor Checklist

### DIFF
--- a/src/zino/scheduler.py
+++ b/src/zino/scheduler.py
@@ -46,7 +46,7 @@ def load_polldevs(polldevs_conf: str) -> Tuple[Set, Set, Set]:
     :returns: A tuple of (new_devices, deleted_devices, changed_devices)
     """
     try:
-        devices = {d.name: d for d in read_polldevs(polldevs_conf)}
+        devices = read_polldevs(polldevs_conf)
     except InvalidConfiguration as error:
         _log.error(error)
         return set(), set(), set()
--- a/src/zino/scheduler.py
+++ b/src/zino/scheduler.py
@@ -81,10 +81,11 @@ def init_state_for_devices(devices: Sequence[PollDevice]):
 async def load_and_schedule_polldevs(polldevs_conf: str):
     new_devices, deleted_devices, changed_devices, defaults = load_polldevs(polldevs_conf)
     deschedule_devices(deleted_devices | changed_devices)
-    schedule_devices(new_devices | changed_devices)
+    stagger_interval = defaults.get("interval", DEFAULT_INTERVAL_MINUTES)
+    schedule_devices(new_devices | changed_devices, int(stagger_interval))
 
 
-def schedule_devices(devices: Sequence[str]):
+def schedule_devices(devices: Sequence[str], stagger_interval: int = DEFAULT_INTERVAL_MINUTES):
     devices = sorted((state.polldevs[name] for name in devices), key=operator.attrgetter("priority"), reverse=True)
     if not devices:
         return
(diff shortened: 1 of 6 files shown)


src/zino/scheduler.py
```diff
@@ -79,13 +79,14 @@ def init_state_for_devices(devices: Sequence[PollDevice]):
...
+    new_devices, deleted_devices, changed_devices, defaults = load_polldevs(polldevs_conf)
     deschedule_devices(deleted_devices | changed_devices)
-    schedule_devices(new_devices | changed_devices)
+    default_interval_minutes = defaults.get("interval", DEFAULT_INTERVAL_MINUTES)
+    schedule_devices(new_devices | changed_devices, int(default_interval_minutes))
 
 
-def schedule_devices(devices: Sequence[str]):
-    devices = sorted((state.polldevs[name] for name in devices), key=operator.attrgetter("priority"), reverse=True)
+def schedule_devices(new_devices: Sequence[str], default_interval_minutes: int = DEFAULT_INTERVAL_MINUTES):
```
stveit, line 88: The name of the new parameter `default_interval_minutes`  doesnt really explain what its actually used for, maybe `stagger_interval` is better.

Also, is the parameter renamed back to `new_devices` on purpose? pretty sure this was changed in #333

johannaengland: No, not on purpose, I was rebasing and seemingly missed that one

src/zino/scheduler.py
```diff
@@ -93,8 +94,8 @@ def schedule_devices(devices: Sequence[str]):
 
     scheduler = get_scheduler()
 
-    # Spread poll jobs evenly across the entire default interval
-    stagger_factor = (DEFAULT_INTERVAL_MINUTES * 60) / len(devices)
+    # Spread poll jobs evenly across the entire interval
+    stagger_factor = (stagger_interval * 60) / len(devices)
```
runborg, line 98: for me doing this convertion here feels not logical. Isn't it better to move this outside the schedule_devices function and up to load_and_schedule_polldevs? That way you just input the number of seconds between devices to the schedule_devices function and keep logic outside it.

runborg (changes requested):
Other than the point noted, this patch works as expected.

src/zino/scheduler.py
johannaengland: This was not logic introduced in this pull request, so in order to fix this bug I would suggest creating a new issue where we can discuss where this logic best belongs. 
Because to me it feels at the right place. The function `load_and_schedule_polldevs` should not have to handle details on how to actually schedule the jobs, only that they are being scheduled

runborg (approved):

stveit (approved):

PR #340, opened 2024-08-02
Use ifalias for Planned Maintenance "str" and "regexp" matching

### LINKED ISSUE #297: Possible wrong value used for port matching in planned maintenance
Looking through the OG zino1 code, it uses `$portToLocIfDescr` for matching against ports. I think that might contain `ifalias` values, while `$portToIfDescr` contains `ifdescr`. The PM code right now uses `port.ifdescr` for matching, but maybe it should be changed to `port.ifalias`?

stveit:
## Scope and purpose

Fixes #297

Original zino seems to use ifalias and ifdescr for planned maintenance matching, while intitial PM implementation used only ifdescr.
This PR changes it to use ifalias where applicable instead. 

Original zino code confuses me so wouldn't be a bad idea for someone else to take a look and see if this PR actually makes the correct changes

## Contributor Checklist

### DIFF
--- a/changelog.d/297.fixed.md
+++ b/changelog.d/297.fixed.md
@@ -1,1 +1,1 @@
-Fix PM matching wrongly using ifdescr instead of the correct ifalias
+Match against port alias instead of port description when match_type is "regexp" or "str" for portstate maintenance events. Still matches port description for "intf-regexp".
--- a/changelog.d/297.fixed.md
+++ b/changelog.d/297.fixed.md
@@ -1,0 +1,1 @@
+Fix PM matching wrongly using ifdescr instead of the correct ifalias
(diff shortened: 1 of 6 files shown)


johannaengland:
I would like to have a news fragment for that, because this is a functionality change

johannaengland (approved):

stveit:
Did another look on original code, and seems it uses both ifalias and ifdescr in different situations. I have changed things to reflect that

changelog.d/297.fixed.md
```diff
@@ -0,0 +1 @@
+Fix PM matching wrongly using ifdescr instead of the correct ifalias
```
johannaengland, line 1: Please update this to be more concrete on when `ifalias` is used for matching

johannaengland (approved):

lunkwill42 (approved):

changelog.d/297.fixed.md
lunkwill42: I can grok both versions, so resolving this. As release manager, I reserve the right to rewrite any and all changelog entries as I see fit before committing the compiled changelog :)

PR #341, opened 2024-08-02
Use enums for match types

stveit:
## Scope and purpose

Uses existing enum for different matching type instead of comparing against strings with exact same values as the enums.
Does not change any functionality, just clears up the code

## Contributor Checklist


johannaengland (approved):
Added a few more instances where it's nice to use the enums

PR #342, opened 2024-08-02
Log warning if secrets file is readable by other users

### LINKED ISSUE #280: Log warning if `secrets` file is world-readable
The legacy `secrets` file contains clear-text password, and users should take care to ensure it isn't readable by anyone but the user that runs the zino process.   Zino should detect if the permissions are too open and log this as a `WARNING` message.

Cannot remember if Zino 1 does this currently, which might affect whether this should be labelled as post-2.0

johannaengland:
## Scope and purpose

Fixes #280.
I did not add an test that it actually logs the warning, but I manually tested it and added tests for the utils function. 

## Contributor Checklist

### DIFF
--- a/changelog.d/280.added.md
+++ b/changelog.d/280.added.md
@@ -1,0 +1,1 @@
+Log warning if `secrets` file is world-readable
--- a/src/zino/utils.py
+++ b/src/zino/utils.py
@@ -1,5 +1,7 @@
 import asyncio
 import logging
+import os
+import stat
 from functools import wraps
 from ipaddress import ip_address
 from time import time
@@ -63,3 +65,10 @@ def log_time_spent(logger: Union[logging.Logger, str] = __name__, level: int = l
         return wrapper
 
     return actual_decorator
+
+
+def file_is_world_readable(file: str) -> bool:
+    """Returns a boolean value indicating if a file is readable by other users than its owner"""
+    st_mode = getattr(os.stat(path=file), "st_mode", None)
+
+    return bool(st_mode & stat.S_IROTH)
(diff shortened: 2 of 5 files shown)


src/zino/utils.py
```diff
@@ -63,3 +65,12 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return actual_decorator
+
+
+def file_is_readable_by_others(file: str) -> bool:
+    """Returns a boolean value indicating if a file is readable by other users than its owner"""
+    st_mode = getattr(os.stat(path=file), "st_mode", None)
+    if not st_mode:
+        return False
```
lunkwill42, line 74: This branch isn't covered by tests. Not quite sure what this branch is here for?  Is there ever a case where `os.stat()` will return an object that doesn't have the `st_mode` attribute?

changelog.d/280.added.md
```diff
@@ -0,0 +1 @@
+Log warning if secrets file is readable by other users than zino user
```
lunkwill42, line 1: This wording is not 100% in sync with the actual code. 

The code checks if the readable bit is set for OTHER users, it doesn't care who the zino user is.  It also doesn't care if there is some non-zino-user group that can access the file (but that would also complicate the logic).

IMHO, verifying that the file is not world-readable is sufficient for most cases, but the changelog might read as follows instead:

```suggestion
Log warning if `secrets` file is world-readable
```

lunkwill42 (changes requested):

src/zino/utils.py
johannaengland: I was uncertain about that, because I could not find any information about it in the documentation (https://docs.python.org/3/library/os.html#os.stat), but I just checked the source code and there it says that `st_mode` is always present in a result of `os.stat()`.

lunkwill42 (approved):
Looks good to me now, but you have merge conflict with master...

PR #344, opened 2024-08-07
Check actual event type and not PmType

stveit:
## Scope and purpose

Fixes a bug where it compares the type of event to a planned maintenance type, which makes no sense. Instead it should check if the event is an instance of the correct event subclass.

This bug luckily did not actually have any impact on the behaviour, since it checks if `event.type` equals `"portstate"`, which effectively does the same as checking `isinstance(event, PortstateEvent)`, which is why this was not picked up by any tests 

## Contributor Checklist

### DIFF
--- a/changelog.d/344.fixed.md
+++ b/changelog.d/344.fixed.md
@@ -1,1 +1,1 @@
-Fix using wrong value to check if event is correct subclass
+When matching an event to a planned maintenance check that event is of the correct subclass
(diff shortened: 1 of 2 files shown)


changelog.d/344.fixed.md
```diff
@@ -0,0 +1 @@
+Fix using wrong value to check if event is correct subclass
```
johannaengland, line 1: This is not specific enough, from just reading this I have no idea which part of Zino has been changed

```suggestion
When matching an event to a planned maintenance check that event is of the correct subclass
```

johannaengland (approved):

PR #359, opened 2024-08-22
Use pre-commit hook by ruff organisation

johannaengland:
## Scope and purpose

When doing research for #358 I stumbled over this repository and realized that the ruff pre-commit hook we're using is a different one than the one made by the organisation that develops ruff. So I switched to that one. 

## Contributor Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
     rev: 24.8.0
     hooks:
       - id: black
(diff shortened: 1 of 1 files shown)


lunkwill42 (approved):
Not entirely true.  Charlie Marsh is the original author of Ruff, and also the founder of Astral.  The `ruff-pre-commit` repo has been moved from his personal GitHub account to the Astral account, so the two URLs are effectively the same (the former automatically redirects to the latter) :)

Issue #364, opened 2024-08-27
Dump Zino state to temporary file and rename when successful


lunkwill42:
Zino 2.0.0-beta.1 still lacks an important safeguard: When the running state is serialized to disk, the existing state file is overwritten directly.  This means that if Zino crashes or the serialization process bugs out midway, we may end up with a truncated or corrupt state file.

As with legacy Zino, every time the running state is persisted to disk, it should be persisted to a temporary file, and only when successful, this file should be renamed to take the place of the live state file.

lunkwill42:
There may already be some issue where this is happening when a lot of polling is going on: The Pydantic serialization process crashes because the data structure it is serializing is changing under its nose.

I can't recall exactly, byt the dumper job may actually be running in a separate thread, which is not a good choice when another thread is accessing and updating the running state in parallel.  Either the dump has to take place in the main thread, blocking everything while it is running, or a snapshot copy of the running state needs to be made somehow before the serialization thread runs, so that it can serialize its own private copy of the state.

PR #365, opened 2024-08-30
Use copy of state for state dumps to file

### LINKED ISSUE #468: Zino sometimes logs a scary looking `PanicException`
Sometimes, users will find that Zino logs this scary looking exception:

```pytb
  File "/home/zino/.zino/lib/python3.11/site-packages/apscheduler/executors/base.py", line 131, in run_job
    retval = job.func(*job.args, **job.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zino/.zino/lib/python3.11/site-packages/zino/utils.py", line 33, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/zino/.zino/lib/python3.11/site-packages/zino/state.py", line 48, in dump_state_to_file
    statefile.write(self.model_dump_json(exclude_none=True, indent=2))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zino/.zino/lib/python3.11/site-packages/pydantic/main.py", line 519, in model_dump_json
    return self.__pydantic_serializer__.to_json(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: dictionary changed size during iteration
```

This is most likely because serialization and dumping of the running state to disk takes place in a separate thread in order not to block the router polling - however, since there are no locking mechanisms in place, the main thread may be modifying the data structures as Pydantic is trying to serialize them to JSON data.

This problem is somewhat mitigated by #364, which ensures the state file is not corrupted if this error occurs. Committing the state to disk will, however, be delayed by this error until the next time a dump is scheduled.

This problem was mentioned in #364 and #365 tries to solve it.  This issue is here to document what the error looks like, so users who experience it can find this issue.

johannaengland:
## Scope and purpose

Mentioned in #364. Fixes #468.

## Contributor Checklist

### DIFF
--- a/src/zino/state.py
+++ b/src/zino/state.py
@@ -43,9 +43,10 @@ class ZinoState(BaseModel):
     def dump_state_to_file(self, filename: str):
         """Dumps the full state to a file in JSON format"""
         _log.debug("dumping state to %s", filename)
+        copied_state = self.model_copy()
         temp_file = f"{filename}.tmp"
         with open(temp_file, "w") as statefile:
-            statefile.write(self.model_dump_json(exclude_none=True, indent=2))
+            statefile.write(copied_state.model_dump_json(exclude_none=True, indent=2))
         os.replace(src=temp_file, dst=filename)
 
     @classmethod
--- a/changelog.d/+use-copy-for-statedump.fixed.md
+++ b/changelog.d/+use-copy-for-statedump.fixed.md
@@ -1,0 +1,1 @@
+Use a copy of the Zino state for saving it to a file to avoid errors due to changes to the state in the middle of saving


src/zino/state.py
```diff
@@ -42,8 +42,9 @@ class ZinoState(BaseModel):
     def dump_state_to_file(self, filename: str):
         """Dumps the full state to a file in JSON format"""
         _log.debug("dumping state to %s", filename)
+        copied_state = self.model_copy()
```
lunkwill42, line 45: I think you may actually want a deep copy of the state to ensure nothing is being updated from another thread:

https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_copy

lunkwill42 (changes requested):
Yes, this looks like the right approach, though I think you probably want a deep copy.

Also not sure how expensive an operation it is for PyDantic to copy the entire structure, which can grow quite large.  If running against the entire set of Uninett routers, is there a noticable difference in the runtime of this function before and after this PR?

src/zino/state.py
johannaengland: If I try to use deep copy I get the following exception
```
2024-09-03 09:48:07,045 - ERROR - apscheduler.executors.default (asyncio_0) - Job "ZinoState.dump_state_to_file (trigger: interval[0:05:00], next run at: 2024-09-03 09:53:07 CEST)" raised an exception
  File "/home/johanna/zino/.venv/lib/python3.10/site-packages/apscheduler/executors/base.py", line 125, in run_job
    retval = job.func(*job.args, **job.kwargs)
  File "/home/johanna/zino/src/zino/utils.py", line 33, in wrapper
    result = func(*args, **kwargs)
  File "/home/johanna/zino/src/zino/state.py", line 45, in dump_state_to_file
    copied_state = self.model_copy(deep=True)
  File "/home/johanna/zino/.venv/lib/python3.10/site-packages/pydantic/main.py", line 311, in model_copy
    copied = self.__deepcopy__() if deep else self.__copy__()
  File "/home/johanna/zino/.venv/lib/python3.10/site-packages/pydantic/main.py", line 777, in __deepcopy__
    _object_setattr(m, '__dict__', deepcopy(self.__dict__, memo=memo))
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.10/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "/home/johanna/zino/.venv/lib/python3.10/site-packages/pydantic/main.py", line 789, in __deepcopy__
    deepcopy({k: v for k, v in self.__pydantic_private__.items() if v is not PydanticUndefined}, memo=memo),
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 206, in _deepcopy_list
    append(deepcopy(a, memo))
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 238, in _deepcopy_method
    return type(x)(x.__func__, deepcopy(x.__self__, memo))
  File "/usr/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.10/copy.py", line 271, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.10/copy.py", line 271, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 206, in _deepcopy_list
    append(deepcopy(a, memo))
  File "/usr/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.10/copy.py", line 271, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 211, in _deepcopy_tuple
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.10/copy.py", line 211, in <listcomp>
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.10/copy.py", line 211, in _deepcopy_tuple
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.10/copy.py", line 211, in <listcomp>
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.10/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle '_asyncio.Future' object
```

lunkwill42: Oooh, that seems pretty weird... what could be putting a `Future` object into this structure?

johannaengland: I'm investigating it - it is not `FlappingStates` - I checked out a commit before it was added and the same error happens

johannaengland: Same with `PlannedMaintenances`

johannaengland: Alright, I have figured out that the problem lies with `ZinoState.events` - I will investigate further which of them lead to problems

lunkwill42: But then again, I'm not sure that this actually solves the issue at hand.  I still get the occasional error about dicts changing size while Pydantic is iterating them to copy them, which was the reason for doing a copy to begin with.  And, of course, this makes sense.  The copy is being made in the same thread that is doing the dumping, so if the structure is being changed by another thread, the same problem will still occur.

The only way to ensure we don't get this error is to ensure that no other threads are concurrently modifying the data structure while it's being serialized to disk (or while it's being copied).  This can be done by either making sure we're not running multiple threads, or by adding locks (which will complicate things quite a bit).

It could also be "worked around", by simply retrying the serialization several times if it fails due to these errors - which would be safe enough when #366 is merged, as that would protect us from leaving a corrupt state file when the error inevitably occurs.

lunkwill42: I do believe the reason the dumping is happening in a separate thread is because the method `dump_state_to_file()` is not declared as async.  When this function is scheduled to run using APscheduler, apscheduler will automatically defer its execution to a separate worker thread because it is a non-async function and would block the entire event loop.

I think the safest way to run the function would be  to declare it as an `async` function, which lets it run in the main event loop thread.  It will, of course, block the event loop while serializing - which could be ok if it's fast, disastrous if it is slow.  A more complicated implementation would be to implement it as a function that copies the state in the main thread, and then defers the actual serialization of this copy to a worker thread (since the disk i/o is what would block the most).

lunkwill42:
I took the liberty of rebasing this onto the latest master and force-pushing it.

PR #367, opened 2024-09-02
Suppress initial interface down alarms with config option

### LINKED ISSUE #257: Suppress alarms/tickets on initial import of a device / interface
When a new device is added to zino without any prior nown state is added to zino the server will generate tickets on all interfaces in down state. 
These alarms are 99% of time alarms that we do not need and on a new instance of zino will this generate thousands of alarm that the user needs to close.

There should be a way to suppress these alarms on the initial run as they are usually not needed.

johannaengland:
## Scope and purpose

Closes #257. 

This is how I interpreted this issue: We have a config option that if it is set to true the first time a `PortStateEvent` would be created with the interface down state nothing happens instead.

## Contributor Checklist

### DIFF
--- a/changelog.d/257.added.md
+++ b/changelog.d/257.added.md
@@ -1,0 +1,1 @@
+Add config option to suppress initial interface down events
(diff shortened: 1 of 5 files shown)


johannaengland:
Settings this to draft/blocked until we get more concrete specification on how this should work

PR #389, opened 2025-01-21
Fix unintended breakdown of MIB object sub-indexes into symbolic representations

lunkwill42:
## Scope and purpose

This fixes a problem discovered in the Zino SNMP abstraction layer during the construction of an alternative SNMP abstraction layer: The `_convert_varbind()` function was intended to break down an object identity into a symbolic name and a numeric index (as represented by an `Identifier` object).

However, because of a misunderstanding of how PySNMP works, and a slight bug in the implementation of the Zino `OID` class, it turns out that the sub-indexes are also converted to symbolic representations if the MIB objects are defined that way - and the OID class appears to have allowed these non-integer elements to be represented within it.

This was not understood as unintended behavior when the BGP state monitor task was authored, causing workarounds for the strange behavior to be written in that plugin, rather than the bug to be reported :-)

It may actually be in our interest to have some MIB-assisted symbolic breakdown of object identities, but the PySNMP objects should not leak through our abstraction layer (which is slated to be replaced), so this is something we might want to look into at a later stage.

### This pull request

* Fixes a bug in the `OID` class.
* Unrolls symbolic index objects back into numeric OIDs.  It might not be an optimal solution, but the only other one I could think of was to translate the object back into an OID and then strip that prefix from the full object identity being translated.  Both methods are kind of ugly.
* Removes the now broken and obsolete workarounds from the BGP state monitor task.

## Contributor Checklist

### DIFF
--- a/src/zino/snmp.py
+++ b/src/zino/snmp.py
@@ -5,7 +5,6 @@ import os
 import threading
 from collections import defaultdict
 from dataclasses import dataclass
-from functools import reduce
 from ipaddress import ip_address
 from typing import Any, NamedTuple, Sequence, Tuple, Union
 
@@ -502,22 +501,13 @@ def _convert_varbind(ident: ObjectIdentity, value: ObjectType) -> SNMPVarBind:
     """Converts a PySNMP varbind pair to an Identifier/value pair"""
     mib, obj, indices = ident.getMibSymbol()
     value = _mib_value_to_python(value)
-    row_index = reduce(lambda acc, index: acc + _index_to_tuple(index), indices, ())
-    return Identifier(mib, obj, OID(row_index)), value
 
+    prefix = SNMP._oid_to_object_type(mib, obj)
+    SNMP._resolve_object(prefix)
+    prefix = OID(prefix[0])
+    row_index = OID(ident).strip_prefix(prefix)
 
-def _index_to_tuple(index: ObjectIdentity) -> Tuple[int, ...]:
-    """Unrolls a PySNMP index object to a tuple of integers"""
-    try:
-        return tuple(index)
-    except TypeError:
-        try:
-            return index.asTuple()
-        except AttributeError:
-            try:
-                return index.getOid()
-            except AttributeError:
-                return (int(index),)
+    return Identifier(mib, obj, row_index), value
 
 
 def _mib_value_to_python(value: SupportedTypes) -> Union[str, int, OID]:
--- a/src/zino/tasks/bgpstatemonitortask.py
+++ b/src/zino/tasks/bgpstatemonitortask.py
@@ -158,7 +158,8 @@ class BGPStateMonitorTask(Task):
             return None
 
         for oid, result in cisco_bgp_info.items():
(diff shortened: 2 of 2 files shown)


src/zino/snmp.py
```diff
@@ -499,11 +500,26 @@ def udp_transport_target(self) -> Union[UdpTransportTarget, Udp6TransportTarget]
...
+    except TypeError:
+        try:
+            return index.asTuple()
+        except AttributeError:
+            try:
+                return index.getOid()
+            except AttributeError:
+                return (int(index),)
+
+
```
lunkwill42, line 522: Upon testing the new SNMP framework I've found that this still isn't up to snuff, as some bits keep getting cut off.  Notably, an object like ` .1.3.6.1.4.1.9.9.187.1.2.5.1.4.1.4.10.0.0.1` would normally be broken down to `CISCO-BGP4-MIB::cbgpPeer2AdminStatus.ipv4."10.0.0.1"`, but what we want from our function is something more like `CISCO-BGP4-MIB::cbgpPeer2AdminStatus.1.4.10.0.0.1`.   I.e. the index returned should be `1.4.10.0.0.1` (`1` is the type of address represented, `4` is the number of subsequent octets that represent the address) - instead this only return `4.10.0.0.1`.  Only the `bgpstatemonitortask` cares right now, but this PR makes it expect the latter.

lunkwill42: I've pushed a new commit where the strategy becomes more or less the same as in the netsnmp-cffi replacement branch: Reconstruct a prefix from the mib and object name, and then strip that from the incoming OID argument to get the row_index suffix value.

src/zino/tasks/bgpstatemonitortask.py
```diff
@@ -158,7 +158,8 @@ async def _get_cisco_bgp_info(self) -> Optional[list[BaseBGPRow]]:
             return None
 
         for oid, result in cisco_bgp_info.items():
-            result["cbgpPeer2RemoteAddr"] = ip_address(oid)
+            _addr_type, _addr_len, addr = oid[0], oid[1], oid[2:]
+            result["cbgpPeer2RemoteAddr"] = ip_address(bytes(addr))
```
stveit, line 162: Is it useful to assign oid[0] and oid[1]  to variables when they are never used? Could simplify with
```suggestion
            result["cbgpPeer2RemoteAddr"] = ip_address(bytes(oid[2:]))
```
Or
```suggestion
            addr = oid[2:]
            result["cbgpPeer2RemoteAddr"] = ip_address(bytes(addr))
```

stveit (changes requested):
commentados

src/zino/tasks/bgpstatemonitortask.py
lunkwill42: I've left them in there on purpose as an explanatory note for future poor souls who might wonder WTH this is picking the magical `[2:]` part of the OID as the address (the preceding underscore of variable names is just the Pythonic way of indicating that this is a throwaway variable)

stveit (approved):

PR #394, opened 2025-02-14
 Add netsnmp-cffi as alternative SNMP polling back-end

lunkwill42:
## Scope and purpose

Fixes the polling bits of #383 (replaces #391)

This refactors the SNMP agent back-end to be even more amenable to replacement, and moves on to add an alternative implementation based on the new [netsnmp-cffi](https://github.com/Uninett/netsnmp-cffi) library.

It does not (currently) replace the `trapd` back-end, which continues to work alongside these changese.  This is potentially a separate PR, but the drawback is currently that the two back-ends use a different set of MIB files (the original PySNMP back-end vendors MIB files translated to Python code, while this back-end introduces a directory of vendored MIB files in their *original* format).

## Contributor Checklist


.sonarcloud.properties
```diff
@@ -10,4 +10,4 @@ sonar.sources=src
 sonar.python.version=3.9, 3.10, 3.11
 sonar.tests=tests
 
-sonar.exclusions=src/zino/mibdumps/**
+sonar.exclusions=src/zino/snmp/mibdumps/**
```
johannaengland, line 13: You should also add the new `mibs` directory to it

pyproject.toml
```diff
@@ -94,7 +95,7 @@ force-exclude = '''
     | buck-out
     | build
     | dist
-    | src/zino/mibdumps
+    | src/zino/snmp/mibdumps
```
johannaengland, line 98:
```suggestion
    | src/zino/snmp/mibdumps
    | src/zino/snmp/mibs
```

johannaengland: And also ignore that one for ruff

johannaengland (changes requested):
Only tiny changes needed

And you have a typo in the commit message of 3fbffdcf61c5c5d0367924d00d0f5e9aa6ee1ab0 (i instead of is)

pyproject.toml
lunkwill42: That would be useless, both for ruff and black.  The new directory contains MIB modules in ASN format and would never be looked at by Python linter tools.

johannaengland: Ah, I wrongly assumed that these would also be `.py` files, my bad

johannaengland (approved):

lunkwill42:
> And you have a typo in the commit message of [3fbffdc](https://github.com/Uninett/zino/commit/3fbffdcf61c5c5d0367924d00d0f5e9aa6ee1ab0) (i instead of is)

That doesn't matter much. This commit will be completely rewritten before taking this out of draft mode, since everything written in that temporary change is obsolete by now.

lunkwill42:
Since this replaced #391, I'm re-uploading this comment from January 27th:

### Initial comparisons of PySNMP-based (`master`) Zino vs. Net-SNMP-based (this PR) Zino

I ran Zino for exactly 10 minutes with a version of `polldevs.cf` that contained **222** routers from the Uninett research network.  

Preliminary conclusion is that the time spent developing `netsnmp-cffi` was worthwhile, as it provides a significant performance improvement (bugs notwithstanding)

#### Resource usage for PySNMP version
`494.50s user 2.64s system 82% cpu 10:00.02 total`
#### Resource usage for Net-SNMP version
`78.03s user 4.94s system 13% cpu 10:00.03 total`

src/zino/config/models.py
```diff
@@ -42,7 +42,7 @@ class PollDevice(BaseModel):
     retries: int = 3
     domain: str = None
     statistics: bool = True
-    hcounters: bool = False
+    hcounters: bool = True  # basically default to SNMP v2c
```
hmpf, line 45: Important comment. Worth documenting?

hmpf (commented):
See question

src/zino/config/models.py
lunkwill42: Zino lacks proper documentation, which is a task of its own - so not sure where I would document this until a docs framework is in place.

I have suggested to Håvard that this option name should change.  It only makes sense when using Zino 1 for statistics collection.  High capacity counters are irrelevant for Zino 2, which doesn't provide a statistics collection replacement, so the option name makes no sense.

I.e. for now, the option name must remain the same in order to keep compatibility with old `polldevs.cf` files, but we should make an issue for replacing/renaming/aliasing it.

johannaengland: #397

tox.ini
```diff
@@ -22,6 +22,7 @@ deps =
     pytest-timeout
     snmpsim>=1.0
     retry
+    netsnmp-cffi>=0.1.1
```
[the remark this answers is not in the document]
lunkwill42: Because tox would not work without it, of course.

However, that was while still in draft mode and this was referencing a GitLab URL, which is no longer the case.

hmpf (approved):

PR #399, opened 2025-03-12
Remove list of development tools from list of optional runtime requirements

lunkwill42:
## Scope and purpose

Optional dependencies in `pyproject.toml` are taken to mean "optional *runtime* dependencies" by Python build and packaging tools.  The list of `dev` optional dependencies are definitely *not runtime* dependencies in any way, shape or form, so this removes them from package metadata and relegates them to a simple requirements file.

## Contributor Checklist

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -12,7 +12,7 @@ pytest-asyncio<0.22.0
 pytest-timeout
 retry
 ruff
-snmpsim>=1.0
+snmpsim>=1.0,!=1.1.6
 towncrier
 tox<4
 twine
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,0 +1,18 @@
+# Development "requirements".  These are tools that are extremely useful or
+# even required to run the test suite, but not at all required by Zino at
+# runtime:
+black
+build
+coverage
+ipython
+isort
+pre-commit
+pytest
+pytest-asyncio<0.22.0
+pytest-timeout
+retry
+ruff
+snmpsim>=1.0
+towncrier
+tox<4
+twine
--- a/README.md
+++ b/README.md
@@ -300,6 +300,17 @@ zinoconv save-state.tcl zino-state.json
 
 ## Developing Zino
 
+### Development tools
+
+A bunch of tools needed or recommended to have available when developing Zino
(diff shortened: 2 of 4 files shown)


hmpf (commented):
*Why* are you doing this? Is it because pip cannot install just optional dependencies?

requirements/dev.txt
```diff
@@ -0,0 +1,18 @@
...
+coverage
+ipython
+isort
+pre-commit
+pytest
+pytest-asyncio<0.22.0
+pytest-timeout
+retry
+ruff
+snmpsim>=1.0
```
johannaengland, line 15:
```suggestion
snmpsim>=1.0,!=1.1.6
```

See #370

johannaengland (changes requested):

requirements/dev.txt
lunkwill42: Strange, I copied the list from `pyproject.toml`, and that restriction was there at the time. No idea why it disappeared...

lunkwill42:
> _Why_ are you doing this? Is it because pip cannot install just optional dependencies?

Two reasons:

1. Optional dependencies (as per `pyproject.toml` standards) are essentially a way to signal dependencies that your package will function without, but will give your package more functionality at runtime if installed.  That is not the case with these dependencies: They have no bearing on runtime functionality.
 
2. NetBSD tools Håvard uses to build NetBSD packages from a Python package will follow said standard, and suggest that all these dependencies must be packaged for NetBSD in order to make a fully functional NetBSD port of the package, complete with optional dependencies available.  Whis is also patently false.

@hmpf: I believe this is the proposed standard we should be using instead, but it's not clear to me that all current build tools support it:  https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups

requirements/dev.txt
lunkwill42: Nope, #370 only updated `tox.ini`, not `pyproject.toml`.  But I just rebased this PR and added a commit to include the same constraint in `dev.txt`

johannaengland (approved):

hmpf:
Looks like PEP 735 is supported by:

* uv (merged, not in release yet)
* pip 25.1 (to come)

PR #400, opened 2025-03-13
Fix Net-SNMP resource leakage issues

lunkwill42:
## Scope and purpose

After #394 and friends were merged, running Zino with the 222 routers of the Uninett research network revealed that Zino was leaking low-level file descriptors.

This adds more in-depth debug logging of how SNMP sessions are re-used and garbage collected, so that the issue can be properly debugged.

It also fixes the inherent leakage problem by adding more control of resource management to the `SNMP` implementations.  See commit logs for more details.

## Contributor Checklist

### DIFF
--- a/src/zino/snmp/netsnmpy_backend.py
+++ b/src/zino/snmp/netsnmpy_backend.py
@@ -79,6 +79,10 @@ class SNMP:
         )
         self.session.open()
 
+    def __del__(self):
+        """Ensures Net-SNMP resources are properly released when sessions are garbage collected"""
+        self.session.close()
+
     async def get(self, *oid: str) -> MibObject:
         """SNMP-GETs the given oid
         Example usage:
--- a/src/zino/snmp/netsnmpy_backend.py
+++ b/src/zino/snmp/netsnmpy_backend.py
@@ -65,7 +65,20 @@ def init_backend():
 
 
 class SNMP:
-    """Represents an SNMP management session for a single device"""
+    """Represents an SNMP management session for a single device.
+
+    Net-SNMP allocates low-level resources for each session, so it is important to close the session when done with
+    it.  The initial Zino 2 codebase did not expect to have to manage low-level resources, so all SNMP operations of
+    this class will implicitly open an unopened or closed session - which means that some parts of Zino *will* have
+    to manage resource de-allocation.
+
+    Complete job runs should manage the session as a context manager, but tasks that schedule SNMP operations outside
+    of the main job loop should be careful to also manage their own resources.
+
+    Example context manager usage:
+    >>> with SNMP(device) as snmp:
+    >>>     uptime = await snmp.get("SNMPv2-MIB", "sysUpTime", 0)
+    """
 
     def __init__(self, device: PollDevice):
         self.device = device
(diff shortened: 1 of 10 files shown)


lunkwill42:
Additionally, I've figured out why `zino.snmp.netsnmpy_backend.SNMP.__del__()` was needed in the first place: I assumed `netsnmpy.session.SNMPSession.__del__()` would do the trick, but it turns out that the `SNMPSession` class of `netsnmp-cffi` is constructed in such a way that no `SNMPSession` objects will be garbage collected until they are explicitly closed - since the class keeps a map of active session objects - thereby keeping them alive indefinitely unless explicitly closed.

So the `__del__()` issue *could* easily be fixed in `netsnmp-cffi`, where it belongs: Except, if I *do* fix it there, and remove the Zino fix, the test suite problem still appears for Zino.

lunkwill42:
Doing low-level resource management with `__del__()` is probably bad™. While Zino was not geared to do low-level resource management for SNMP sessions, it seems to be necessary for safe and leakage-free operation with the Net-SNMP library.  Adding `__del__()` seemed to work for a running Zino, but the test suite went haywire, and it was near impossible to debug.  My conclusion is that we should be more explicit about how SNMP resources are managed, and that the PySNMP implementation, that doesn't need it, must provide faux methods to do it (for however long we let that implementation live on).

src/zino/snmp/netsnmpy_backend.py
```diff
@@ -79,6 +79,10 @@ def __init__(self, device: PollDevice):
         )
         self.session.open()
 
+    def __del__(self):
+        """Ensures Net-SNMP resources are properly released when sessions are garbage collected"""
+        self.session.close()
+
```
johannaengland: Do we need this commit if we end up replacing it anyway?

tests/zino_test.py
```diff
@@ -269,3 +269,13 @@ def test_when_pysnmp_is_backend_it_should_not_log_low_level_details(self, state_
...
+
+    def test_when_debug_logging_is_not_enabled_it_should_not_log_anything(self, state_with_localhost, caplog):
+        import zino.snmp.netsnmpy_backend as backend
+
+        with patch.object(zino.state, attribute="state", new=state_with_localhost):
+            with patch.object(zino, attribute="import_snmp_backend") as import_snmp_backend:
+                import_snmp_backend.return_value = backend
+                with caplog.at_level(logging.INFO):
+                    zino.log_snmp_session_stats()
+                    assert "gc reachable" not in caplog.text
```
johannaengland, line 281: Shouldn't this be in the commit that introduces the logging?

johannaengland (commented):
Looks good, only comments about commit history

src/zino/snmp/netsnmpy_backend.py
lunkwill42: I made a choice to keep it so that the history will show that this was a dead-end solution (and potentially, maybe, serve as a warning to not try it again ;-) )

tests/zino_test.py
lunkwill42: Yes, this must be a fixup commit gone wrong :P

src/zino/snmp/netsnmpy_backend.py
johannaengland: Good, I approve of that, just wanted to make sure it was intentional

johannaengland (approved):

PR #403, opened 2025-03-19
Use ruff for formatting and linting (as replacement for black and isort)

johannaengland:
## Scope and purpose

This replaces the use of black, isort and MegaLinter by just using ruff for everything.

To get ruff to fix isort rules automatically it is necessary to add the argument `--fix` to the pre-commit hook and add `fixable=["I"] to the ruff linting configuration.  

### This pull request
* removes a dev-dependency

## Contributor Checklist

  - devex
  - devex, but tested manually that isort rules are automatically fixed by ruff, but nothing else


lunkwill42 (changes requested):
Looks ok to me, but since you're reformatting the codebase using ruff's slightly different-to-black rules, this is missing the addition of a `.git-blame-ignore-revs` file (and doc instructions on how to use it) :)

johannaengland:
> Looks ok to me, but since you're reformatting the codebase using ruff's slightly different-to-black rules, this is missing the addition of a `.git-blame-ignore-revs` file (and doc instructions on how to use it) :)

Yes, I did remember that, I just thought in case this needed to be rebased before merging I would push it directly to master after merging this in, but I can also add it and change it in that case

pyproject.toml
```diff
@@ -65,42 +65,16 @@ exclude = ["tests*"]
...
-    | \.ve
-    | _build
-    | buck-out
-    | build
-    | dist
-    | src/zino/snmp/mibdumps
-    | docs
-  )
-)
-'''
```
hmpf, line 90: Does ruff implicitly ignore these?

johannaengland: Yes, it automatically ignores everything in `.gitignore` (https://docs.astral.sh/ruff/settings/#respect-gitignore) and others are also automatically ignored: https://docs.astral.sh/ruff/settings/#exclude

lunkwill42: @johannaengland how about `src/zino/snmp/mibdumps` ?

johannaengland: That is already ignored: https://github.com/Uninett/zino/blob/31f8b41f4ac16180ced21d5db7fe9cb6db2ec7e7/pyproject.toml#L95-L97

But I just noticed that I should use `extend_exclude` there, not `exclude`

lunkwill42 (approved):
Nice work :)

PR #437, opened 2025-09-15
Use new `snmpversion` config field to decide SNMP version

### LINKED ISSUE #397: Rename, replace or alias `hcounters` option in `polldevs.cf` to something more sensible
> I have suggested to Håvard that this option name should change.  It only makes sense when using Zino 1 for statistics collection.  High capacity counters are irrelevant for Zino 2, which doesn't provide a statistics collection replacement, so the option name makes no sense.
>
> I.e. for now, the option name must remain the same in order to keep compatibility with old `polldevs.cf` files, but we should make an issue for replacing/renaming/aliasing it.

_Originally posted by @lunkwill42 in https://github.com/Uninett/zino/pull/394#discussion_r1991036865_

stveit:
## Scope and purpose

Fixes #397.

Replaces hcounters for this purpose, but hcounters is kept as a valid value in the config so old configs that uses hcounters doesnt cause it to crash.

## Contributor Checklist

* [x] Added/changed documentation (Changed example config file)

### DIFF
--- a/changelog.d/397.changed.md
+++ b/changelog.d/397.changed.md
@@ -1,0 +1,1 @@
+Use new snmpversion config field to decide which SNMP version to use.
--- a/polldevs.cf.example
+++ b/polldevs.cf.example
@@ -11,6 +11,7 @@
 default interval: 5
 default community: public
 default domain: example.org
+default snmpversion: v2c
 
 # Example router, with a higher scheduling priority.  Interfaces whose name
 # matches the regular expression `pe-` will be not be monitored:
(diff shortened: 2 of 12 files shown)


lunkwill42 (changes requested):
LGTM.  

- Moved from deployment project to current sprint board.  (The deployment project is only to keep track of the status of the issues outward to NORDUnet).
- Only thing I would nitpick is maybe add a test that `hcounters` attributes are *indeed accepted* in the config file.

Test suite is acting weird all over on GitHub Actions right now. We should investigate outside the context of this PR....

PR #441, opened 2025-09-17
Format and test only code on github

hmpf:
## Scope and purpose

Speed up github actions by not testing docs-changes.

## Contributor Checklist

* ~[ ] Added a changelog fragment for [towncrier](https://github.com/Uninett/zino/blob/master/README.md#before-merging-a-pull-request)~
* ~[ ] Added/amended tests for new/changed code ~
* ~[ ] Added/changed documentation~
* ~[ ] Linted/formatted the code with ruff, easiest by using [pre-commit](https://github.com/Uninett/zino/blob/master/README.md#code-style)~

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,7 +3,17 @@ name: Verify formatting
 on:
   push:
     branches: master
+    paths:
+      - '**.py'
+      - 'src/**'
+      - 'tests/**'
+      - '!changelog.d/*'
   pull_request:
+    paths:
+      - '**.py'
+      - 'src/**'
+      - 'tests/**'
+      - '!changelog.d/*'
 
 jobs:
   ruff:
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,17 +3,15 @@ name: Verify formatting
 on:
   push:
     branches: master
-    paths:
-      - '**.py'
-      - 'src/**'
-      - 'tests/**'
-      - '!changelog.d/*'
+    paths-ignore:
+      - 'docs/**'
+      - 'changelog.d/*'
+      - 'README.md'
   pull_request:
-    paths:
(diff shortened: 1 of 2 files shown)


lunkwill42 (changes requested):
This is a good idea (though not extremely important for Zino, whose test suite is quite fast compared to other projects we work on).

However, I'm not 100% convinced that the paths are foolproof. We might touch stuff that isn't Python code, or located in the `src` or `test` directories, and it would still affect tests.  One concrete example would be changing dependencies or other project details in `pyproject.toml` - this would definitely affect tests.  Other examples can probably also be made.

I'm not sure what tests we would have that target code files from the doc tree - wouldn't ignoring a workflow that only touches files in the doc tree suffice?

(reference for other readers: 
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore )

hmpf:
> I'm not sure what tests we would have that target code files from the doc tree - wouldn't ignoring a workflow that only touches files in the doc tree suffice?

The idea was: if we ever have an example python file that is only in the docs-tree. With the literalinclude role we shouldn't need that though.

johannaengland:
I agree with @lunkwill42 that there are probably a lot of changes that should have the test suite running that are not included in this (another thing that I am think of are changes to `test.yml`.

I think we are safer just excluding changes to docs, `changelog.md/` and the README

hmpf:
Eh.., just ignoring means full tests are run for *this* PR thus no longer demonstrating that the filter actually works =D

johannaengland:
> Eh.., just ignoring means full tests are run for _this_ PR thus no longer demonstrating that the filter actually works =D

But that is good, since changes to how the tests are run on Github should check that the tests are still running afterwards (don't ask me how I know :sweat_smile:)

PR #451, opened 2025-09-18
Use fullstring matching in `PlannedMaintenance`

### LINKED ISSUE #450: Use fullstring matching for regexp matches in `PlannedMaintenance`
Similar to #426: Original zino uses fullstring matching for regexp matching, while the current Zino2 implementation only matches on start of string.

The fix is to change `re.match` to `re.search`

stveit:
## Scope and purpose

Fixes #450 

Changes `re.match` to `re.search `so it uses uses fullstring matching.
updates existing tests and adds new ones to cover this

## Contributor Checklist

* [ ] The first line of the commit message continues the sentence "If applied, this commit will ...", starts with a capital letter, does not end with punctuation and is 50 characters or less long. See https://cbea.ms/git-commit/

### DIFF
--- a/changelog.d/450.fixed.md
+++ b/changelog.d/450.fixed.md
@@ -1,0 +1,1 @@
+Use fullstring matching for regexp matches in PlannedMaintenance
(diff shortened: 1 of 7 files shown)


lunkwill42 (approved):
LGTM

PR #466, opened 2025-09-22
Use correct configuration object in `main()`

### LINKED ISSUE #458: Zino cannot run without `zino.toml` file
When having added a secrets file, but no `zino.toml` file the following error occurs:

```
2025-09-19 12:48:16,336 - WARNING - zino (MainThread) - Secrets file secrets is world-readable. Please ensure that it is only readable by the user that runs the zino process.
  File "/home/johanna/zino/.venv/bin/zino", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/johanna/zino/src/zino/zino.py", line 67, in main
    snmp_backend = import_snmp_backend(config.snmp.backend)
                                       ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'snmp'
```

It should be possible to run Zino without a `zino.toml` file.

lunkwill42:
## Scope and purpose

Fixes #458.

Parts of `main()` was referencing the incorrect config object when loading the configured SNMP back-end. The "official" running config is in the state module. It is only replaced if `load_config()` returns an actual config object.  `load_config()` returns an empty result if the default config file is not present, but raises an exception if an alternate non-existent config file was specified on the command line.

In order for startup without a config file to work, all config must be referenced via `state.config`.

## Contributor Checklist

* [ ] ~~Added/amended tests for new/changed code ~~

### DIFF
--- a/changelog.d/458.fixed.md
+++ b/changelog.d/458.fixed.md
@@ -1,0 +1,1 @@
+Don't crash when default `zino.toml` is not found
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -48,9 +48,9 @@ def main():
         level=logging.INFO if not args.debug else logging.DEBUG,
         format="%(asctime)s - %(levelname)s - %(name)s (%(threadName)s) - %(message)s",
     )
-    config = load_config(args)
-    if config:
-        state.config = config
+    _config = load_config(args)
+    if _config:
+        state.config = _config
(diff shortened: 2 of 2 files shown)


hmpf:
This solves the immediate problem but I'm keeping my branch around because as it is today, with this fix, how zino's config is implemented is hard to understand, hard to explain, hard to debug and unnecessarily hard to test.

PR #475, opened 2025-09-23
Do async reverse_dns call before creating event

stveit:
## Scope and purpose

Creating an event then doing an async call before committing it can cause issues with race conditions if an event with same index is made elsewhere while waiting for the response. Related to #474

## Contributor Checklist

* [ ] Linted/formatted the code with ruff, easiest by using [pre-commit](https://github.com/Uninett/zino/blob/master/README.md#code-style)

### DIFF
--- a/src/zino/tasks/bfdtask.py
+++ b/src/zino/tasks/bfdtask.py
@@ -69,6 +69,10 @@ class BFDTask(Task):
         port.bfd_state = new_state
 
     async def _create_or_update_event(self, port: Port, new_state: BFDState):
+        neigh_rnds = None
+        if new_state.session_addr:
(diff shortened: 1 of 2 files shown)


src/zino/tasks/bfdtask.py
```diff
@@ -69,6 +69,10 @@ async def _update_state(self, port: Port, new_state: BFDState):
         port.bfd_state = new_state
 
     async def _create_or_update_event(self, port: Port, new_state: BFDState):
+        neigh_rnds = None
```
lunkwill42, line 72: ARgh, no-one even noticed the stupid typo...

PR #480, opened 2025-10-01
Change the `Task` base class and the `Zino1BaseServerProtocol` class to accept a reference to a `Configuration` object

### LINKED ISSUE #477: Change the `Task` base class to accept a reference to a `Configuration` object
The `zino.tasks.task.Task` class currently accepts a `state` argument to its `__init__()` method. This argument should be a `ZinoState` object, which the task can use to get or update the running state of the system without needing to reach into global variables:

https://github.com/Uninett/zino/blob/e9678bf5bc7ccdeba009a368480e020416d0cdc7/src/zino/tasks/task.py#L15-L19

This allows us to change how or where the current Zino state is "stored" without needing to update `Task` subclasses, and it also makes writing tests for `Task` subclasses much easier, since any state can be describe and given as input.

Up until the last month, no `Task` subclasses needed to access the running *configuration* of Zino.  #443 was the first to use this, but since the `Task` base class doesn't have a reference to the current config, the code reaches into the global configuration variable `zino.state.config`, which is not a nice(tm) solution.

It would be much better to also provide each `Task` instance with a reference to the currently active `Configuration` object - which improves decoupling of the code and, again, eases the task of writing tests that test configuration-modifiable behavior of `Task` subclasses.

johannaengland:
## Scope and purpose

This was a quite rabbit hole I fell into...

Fixes #477. Dependent on #462.

Also fixed that in https://github.com/Uninett/zino/blob/7c403363065d6bc87079ab619b7f7eeff9f711be/src/zino/api/legacy.py#L22 we imported the configuration on import, which might be a problem if it is imported before the configuration is read from a file.

## Contributor Checklist

  don't think it needs one, since this should not change anything for the user?


lunkwill42:
A rebase is perhaps in order?

tests/tasks/test_task.py
```diff
@@ -16,14 +17,14 @@ async def test_get_sysuptime_returns_uptime(self, snmpsim, snmp_test_port):
...
-        task = ReachableTask(device, state)
+        task = ReachableTask(device, state, zino.state.config)
         uptime = await task._get_uptime()
         assert uptime
 
     async def test_get_sysuptime_raises_timeout_error(self):
         device = PollDevice(name="nonexist", address="127.0.0.1", community="invalid", port=666)
         state = ZinoState()
-        task = ReachableTask(device, state)
+        task = ReachableTask(device, state, zino.state.config)
```
lunkwill42, line 27: Why does this task need the global config variable during tests?  Won't an empty (i.e. default values only) `Configuration` object do?

tests/tasks/test_linkstatetask.py
```diff
@@ -239,15 +240,15 @@ async def test_when_timeout_occurs_poll_single_interface_should_not_crash(self,
 def linkstatetask_with_links_up(snmpsim, snmp_test_port):
     device = PollDevice(name="buick.lab.example.org", address="127.0.0.1", port=snmp_test_port, community="linksup")
     state = ZinoState()
-    task = LinkStateTask(device, state)
+    task = LinkStateTask(device, state, zino.state.config)
```
lunkwill42, line 243: Why does this task need the global config variable during tests?  Won't an empty (i.e. default values only) `Configuration` object do?

tests/tasks/test_linkstatetask.py
```diff
@@ -239,15 +240,15 @@ async def test_when_timeout_occurs_poll_single_interface_should_not_crash(self,
...
     yield task
 
 
 @pytest.fixture
 def linkstatetask_with_one_link_down(snmpsim, snmp_test_port):
     device = PollDevice(name="buick.lab.example.org", address="127.0.0.1", port=snmp_test_port, community="linksdown")
     state = ZinoState()
-    task = LinkStateTask(device, state)
+    task = LinkStateTask(device, state, zino.state.config)
```
lunkwill42, line 251: Why does this task need the global config variable during tests?  Won't an empty (i.e. default values only) `Configuration` object do?

tests/tasks/test_linkstatetask.py
```diff
@@ -257,13 +258,13 @@ def linkstatetask_with_admin_down(snmpsim, snmp_test_port):
         name="buick.lab.example.org", address="127.0.0.1", port=snmp_test_port, community="linksadmindown"
     )
     state = ZinoState()
-    task = LinkStateTask(device, state)
+    task = LinkStateTask(device, state, zino.state.config)
```
lunkwill42, line 261: Why does this task need the global config variable during tests?  Won't an empty (i.e. default values only) `Configuration` object do?

lunkwill42 (changes requested):
Nice work :)

PR #491, opened 2025-11-18
Add and use `Event.closed` timestamp to track event closure times

### LINKED ISSUE #490: Zino 2 incorrectly updates `Event.updated` on event closure
As reported by @he32, Zino 2 seems to be inconsistent with Zino 1 behavior when it comes to closing events:

* Zino 1 will only ever update the timestamp of `Event.updated` when a log entry is added to the event.
* Zino 2 also updates `Event.updated` if an event commit operation changes the administrative state of the Event to `closed`.

Additionally, the event scavenging routine in Zino 2 assumes the  `Event.updated` value of closed events only ever represents the time the event was closed.  This is incorrect, as some flap handling routines may actually make updates to closed events after the fact - which in turn could lead to these events never being expired and evicted from the running state.  Zino 1 tracks event close times in a separate attribute, Zino 2 should replicate this.

lunkwill42:
## Scope and purpose

Fixes #490 by adding the `Event.closed` timestamp attribute (Zino 1 would use the global dictionary `EventCloseTimes` to track this), and updating event expiry routines to calculate expiry times from this attribute (i.e. no code should expect `Event.updated` to represent a closure timestamp for any event).

## Contributor Checklist

### DIFF
--- a/tests/events_test.py
+++ b/tests/events_test.py
@@ -201,6 +201,18 @@ class TestEvents:
         events.commit(event)
         assert event.closed == original_closed
 
+    def test_given_closed_event_without_closed_timestamp_then_delete_expired_events_should_set_closed_timestamp(
+        self, tmp_path
+    ):
+        events = Events()
+        event = events.get_or_create_event("foobar", None, ReachabilityEvent)
+        event.set_state(EventState.CLOSED)
+        events.commit(event)
+        event.closed = None
+        with patch("zino.config.models.EVENT_DUMP_DIR", tmp_path):
+            events.delete_expired_events()
+        assert event.closed <= now()
+
     def test_delete_expired_events_should_delete_old_closed_event(self, tmp_path):
         events = Events()
         event = events.get_or_create_event("foobar", None, ReachabilityEvent)


src/zino/events.py
```diff
@@ -192,9 +192,13 @@ def _delete(self, event: Event):
...
-        event_list = list(self.events.values())
-        for event in event_list:
-            if event.state == EventState.CLOSED and now() > (event.updated + EVENT_EXPIRY):
+        closed_events = list(event for event in self.events.values() if event.state == EventState.CLOSED)
+        for event in closed_events:
+            if not event.closed:
+                # Fixup close time for events that were closed but never had a close time set, to ensure it will
+                # expire eventually
+                event.closed = now()
+            if now() > (event.closed + EVENT_EXPIRY):
```
johannaengland:
```suggestion
            elif now() > (event.closed + EVENT_EXPIRY):
```

Since we know that the others are just set to `now` and therefore don't need to check them
But disregard this if you think the other way is clearer

johannaengland: Good point, did not consider this, then your approach is better!

tests/events_test.py
```diff
@@ -172,15 +172,14 @@ def test_commit_should_add_event_to_closed_index_when_closing_event(self):
         events.commit(event)
         assert events.get_closed_event(*identifiers) == event
 
-    def test_commit_should_set_updated_when_closing_event(self):
+    def test_commit_should_not_set_updated_when_closing_event(self):
```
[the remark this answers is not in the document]
lunkwill42: Also, this great suggestion revealed a new logic bug with the indentation change you questioned: Though I'm not sure it would ever happen in a real-world situation, the test suite will actually attempt to commit events in the closed state without ever having them committed in an open state.  In these cases, the `event.closed` timestamp would never be set, since the event had never been entered into the index of open events.  My latest returns the original indentation, but sets the timestamp only if it has no value already.

johannaengland (approved):

PR #504, opened 2025-11-27
Reduce default SNMP BULK-GET `max-repetitions` value from 10 to 5

lunkwill42:
## Scope and purpose

It seems several Uninett routers have issues with IF-MIB walk operations with a max-repetitions of 10, causing many TimeoutErrors in Zino.  This reduces the default on most operations from 10 to 5, which seems to be safer. Eventually, this should probably become a configuration option in `polldevs.cf`.

## Contributor Checklist

* [ ] ~~Added/amended tests for new/changed code ~~


hmpf (changes requested):
Move the magic number to a constant, with maybe a comment of why it is exactly 5.

PR #524, opened 2026-02-18
Fork child process for state dumps

### LINKED ISSUE #468: Zino sometimes logs a scary looking `PanicException`

lunkwill42:
## Scope and purpose

Fixes #468. If accepted, this makes #365 obsolete.

State serialization (`model_dump_json`) runs in a thread pool by default (in order to avoid blocking the event loop), which can cause intermittent `PanicException`s ("dictionary changed size during iteration") when the main thread modifies state data structures concurrently.

This PR adds `fork_and_dump_state`, an async wrapper that forks a child process to serialize state. The child inherits a copy-on-write memory snapshot, so it can serialize at leisure while the parent continues unblocked. Being a coroutine, `APScheduler` runs it on the event loop thread (not in the thread pool), and `os.fork()` itself is effectively instantaneous.

`dump_state_to_file` is unchanged and remains directly testable.

This serialization trick is inspired by Redis BGSAVE.

## Contributor Checklist

### DIFF
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -179,6 +179,7 @@ async def fork_and_dump_state(filename: str):
         try:
             state.state.dump_state_to_file(filename)
         except Exception:
+            _log.exception("State dump failed in child process")
             os._exit(1)
         os._exit(0)
     else:
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -225,7 +225,7 @@ def setup_initial_job_schedule(loop: AbstractEventLoop, args: argparse.Namespace
     # Schedule state dumping as often as configured in
     # 'config.persistence.period' and reschedule whenever events are committed
     scheduler.add_job(
-        func=state.state.dump_state_to_file,
+        func=fork_and_dump_state,
         trigger="interval",
         args=(state.config.persistence.file,),
         id=STATE_DUMP_JOB_ID,
--- a/src/zino/zino.py
+++ b/src/zino/zino.py
@@ -174,7 +174,7 @@ async def fork_and_dump_state(filename: str):
         return
 
     pid = os.fork()
-    if pid == 0:
+    if pid == 0:  # pragma: no cover
         # Child process — serialize and exit
         try:
             state.state.dump_state_to_file(filename)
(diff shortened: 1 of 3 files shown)


src/zino/zino.py
```diff
@@ -158,6 +160,50 @@ def init_event_loop(args: argparse.Namespace, loop: Optional[AbstractEventLoop]
...
+        _log.warning("Previous state dump (pid %d) still running, skipping", _dump_child_pid)
+        return
+
+    pid = os.fork()
+    if pid == 0:  # pragma: no cover
+        # Child process — serialize and exit
+        try:
+            state.state.dump_state_to_file(filename)
+        except Exception:
+            os._exit(1)
```
hmpf: Log the actual traceback?

johannaengland (approved):

src/zino/zino.py
lunkwill42: Added.

PR #536, opened 2026-04-07
Add straps/nmtrapd trap multiplexer support

### LINKED ISSUE #362: Add support for receiving SNMP traps through a multiplexer
### Rationale

As mentioned in [Håvard Eidnes' comments in #173](https://github.com/Uninett/zino/issues/173#issuecomment-2014972248), there is often a limit to how many trap receivers that can be meaningfully configured in any one router.

For this reason (and other reasons), the legacy Zino does not actually provide its own SNMP trap listener, but instead connects to an SNMP trap multiplexer service (`straps`, or `nmtrapd` in later incarnations, a part of the Scotty framework, I think).  The trap multiplexer receives the actual trap messages, and multiplexes them to a UNIX socket that multiple processes can read from.  Under this model, Zino never has to be started as a privileged user, only the trap multiplexer does.

This model is not currently supported by Zino 2.  It provides its own trap listener, as we wanted it to be as standalone as possible, without introducing dependencies to the older Tcl-based tools.

However, Zino 2 could be extended to also support some kind of trap multiplexer, whether it be `straps`/`nmtrapd` or something else.

### Some implementation considerations

- A trap multiplexer like `nmtrapd` will add an extra header to an SNMP trap message before passing it on to a UNIX socket.  This header typically contains information about the originators IP address and the length of the incoming packet.
- Zino would have to connect to the UNIX socket and get its trap packets from there.  We would have to find some way to feed a raw SNMP trap packet into the `pynetsnmp` framework, rather than having `pynetsnmp` read it from "the wire" itself.
- Zino 1 features a "watchdog" service: Since the connection to the UNIX socket could be lost without warning, the watchdog takes to automatically re-connect to the UNIX socket if a trap has not been received in some specified interval of time.

lunkwill42:
## Scope and purpose

Fixes #362.

Adds optional support for receiving SNMP traps through an external trap multiplexer (straps or nmtrapd) instead of binding directly to a UDP port.  This allows Zino to receive traps without root privileges, matching the mode of operation used by Zino 1 in production.

### This pull request
* Extracts shared trap processing logic into `NetsnmpTrapProcessorMixin`
* Adds `[snmp.trap]` configuration fields: `source`, `straps_socket`, `nmtrapd_host`, `nmtrapd_port`
* Adds `StrapsTrapReceiver` backend with automatic reconnection and watchdog
* Wires up the new backend in `init_event_loop()`

This depends on Uninett/netsnmp-cffi#21 (`parse_raw_trap()`), which has was just released in version 0.2.0 on PyPI.  The `netsnmp-cffi` dependency is therefore updated to require at least the new version.

## Contributor Checklist

* [x] Added a changelog fragment for towncrier
* [x] Linted/formatted the code with ruff
* [x] The first line of the commit message continues the sentence "If applied, this commit will ...", starts with a capital letter, does not end with punctuation and is 50 characters or less long
* [x] ~~If applicable: Created new issues if this PR does not fix the issue completely/there is further work to be done~~

### DIFF
--- a/changelog.d/362.added.md
+++ b/changelog.d/362.added.md
@@ -1,0 +1,1 @@
+Added support for receiving SNMP traps through a straps or nmtrapd trap multiplexer as an alternative to direct UDP binding.
--- a/tests/trapd/test_straps_backend.py
+++ b/tests/trapd/test_straps_backend.py
@@ -1,0 +1,411 @@
+"""Tests for the straps/nmtrapd trap multiplexer backend."""
+
+import asyncio
+import ipaddress
+import struct
+import time
+from unittest.mock import AsyncMock, Mock, patch
+
+import pytest
+import pytest_asyncio
+
+from zino.config.models import TrapConfiguration
+from zino.state import ZinoState
+from zino.trapd.straps_backend import (
+    NmtrapdFrameReader,
+    StrapsFrameReader,
+    StrapsTrapReceiver,
+)
+
+# The same BER-encoded SNMPv2c coldStart trap used in netsnmp-cffi tests
+SNMPV2C_COLDSTART_TRAP = bytes.fromhex(
+    "305402010104067075626c6963a74702034ff374020100020100303a300e06082b060102010103004302"
+    "30393017060a2b06010603010104010006092b0601060301010501300f060a2b060102010202010101"
+    "02012a"
+)
+
+
+class TestStrapsFrameReader:
+    async def test_when_reading_valid_frame_then_it_should_extract_source_address(self):
+        frame = make_straps_frame("10.0.0.1", 162, SNMPV2C_COLDSTART_TRAP)
+        reader = asyncio.StreamReader()
+        reader.feed_data(frame)
+        addr, port, pdu = await StrapsFrameReader().read_frame(reader)
+        assert addr == ipaddress.IPv4Address("10.0.0.1")
+
+    async def test_when_reading_valid_frame_then_it_should_extract_source_port(self):
+        frame = make_straps_frame("10.0.0.1", 4242, SNMPV2C_COLDSTART_TRAP)
+        reader = asyncio.StreamReader()
+        reader.feed_data(frame)
+        addr, port, pdu = await StrapsFrameReader().read_frame(reader)
+        assert port == 4242
+
+    async def test_when_reading_valid_frame_then_it_should_extract_raw_pdu(self):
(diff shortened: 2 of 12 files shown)


lunkwill42:
Squashed my fixup commits, since no-one has reviewed this through GitHub yet (only IRL)

tests/trapd/test_straps_backend.py
```diff
@@ -0,0 +1,408 @@
...
+def make_straps_frame(addr: str, port: int, pdu: bytes) -> bytes:
+    """Builds a straps-formatted frame: addr(4,network) + port(2,network) + length(4,host)."""
+    addr_bytes = ipaddress.IPv4Address(addr).packed
+    return struct.pack("!4sH", addr_bytes, port) + struct.pack("=I", len(pdu)) + pdu
+
+
+def make_nmtrapd_frame(addr: str, port: int, pdu: bytes) -> bytes:
+    """Builds an nmtrapd-formatted frame: version(1) + unused(1) + port(2) + addr(4) + length(4), all network order."""
+    addr_int = int(ipaddress.IPv4Address(addr))
+    return struct.pack("!BBHII", 0, 0, port, addr_int, len(pdu)) + pdu
```
johannaengland: Q: Shouldn't helper functions be at the bottom of the test file?

johannaengland: These could be fixtures instead, but works as well as functions

tests/trapd/test_straps_backend.py
```diff
@@ -0,0 +1,408 @@
...
+        ):
+            receiver._watchdog_task = asyncio.ensure_future(receiver._watchdog_loop())
+            await asyncio.sleep(0.5)
+            receiver.close()
+            await _gather_cancelled(receiver)
+
+        assert not connect_called
+
+    @pytest.mark.timeout(5)
+    async def test_when_reconnect_succeeds_then_backoff_should_return(
```
johannaengland: I'm confused by this test name

changelog.d/362.added.md
```diff
@@ -0,0 +1 @@
+Added support for receiving SNMP traps via a straps/nmtrapd trap multiplexer as an alternative to direct UDP binding.  Configure `[snmp.trap] source = "straps"` in `zino.toml` to connect to a straps UNIX domain socket.
```
johannaengland: Are you mentioning only source "straps" here and not "nmtrapd", because straps is the one that was requested?

johannaengland (changes requested):

johannaengland (approved):

PR #537, opened 2026-04-10
Add `bare-except` rule to ruff

johannaengland:
## Scope and purpose

Inspired by Uninett/nav#3347. Decided to also add it to Zino after a talk with @lunkwill42. 

Ref: https://docs.astral.sh/ruff/rules/blind-except/

I added missing noqas, feel free to push changes if there are any added noqas that can be replaced by explicit exception catches.

## Contributor Checklist

  workflow changes, not visible to users

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ extend-exclude = [
 ]
 
 [tool.ruff.lint]
-select = ["E4", "E7", "E9", "F", "I"]
+select = ["BLE001", "E4", "E7", "E9", "F", "I"]
 fixable = ["I"]
 
 [tool.coverage.report]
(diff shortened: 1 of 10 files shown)


Simrayz (approved):
Good idea to add this, bare exceptions are a code smell and incomplete way of handling errors. 

When are we going to fix the bare exceptions that are exempted from the rule though? Approved given that a follow-up issue is created for fixing the cases you have annotated with `# noqa:`

lunkwill42:
> Good idea to add this, bare exceptions are a code smell and incomplete way of handling errors.
> 
> When are we going to fix the bare exceptions that are exempted from the rule though? Approved given that a follow-up issue is created for fixing the cases you have annotated with `# noqa:`

An actual bare exception handler is when you only have `except:` with no mentioned exception class. Those are never ok. There are valid reasons to have `Exception` handlers, though, and in those *thought-through* scenarios, the only way to avoid a linter complaint is to add a `noqa`.

PR #540, opened 2026-04-30
Pin snmpsim to version that's not 1.2.2

johannaengland:
## Scope and purpose

The new release makes the tests fail since pysmi is now an optional dependency, but it is used in the code. 

Changes in newest release: https://github.com/lextudio/snmpsim/compare/v1.2.1...v1.2.2#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L33

### This pull request
* changes a dependency

## Contributor Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ test = [
     "pytest-cov",
     "pytest-timeout",
     "retry",
-    "snmpsim>=1.0,!=1.1.6,<1.2",
+    "snmpsim>=1.0,!=1.1.6,!=1.2.2",
     "tox",
 ]
 maintenance = [


pyproject.toml
```diff
@@ -54,7 +54,7 @@ test = [
     "pytest-cov",
     "pytest-timeout",
     "retry",
-    "snmpsim>=1.0,!=1.1.6,<1.2",
+    "snmpsim>=1.0,!=1.1.6,!=1.2.2",
```
lunkwill42, line 57: is it necessary here? This was already pinned to a version *less* than 1.2...

lunkwill42 (changes requested):
more request for comment than change :)

pyproject.toml
johannaengland: Version 1.2.1 works, so I think it's good to use newer versions if available

johannaengland: Or looking at the tests, it looks like that for some reason with Zino version 1.2.1 does not work...

PR #544, opened 2026-05-18
Coalesce legacy API response writes per command

lunkwill42:
## Scope and purpose

Fixes #543.

Existing Zino clients (`curitz`, anything built on `zinolib`) were taking many seconds to fetch the list of open cases from Zino 2 over real networks. The legacy API server's per-line `transport.write()` pattern was emitting 10–15 tiny TCP segments per command response, which interacted badly with Nagle + TCP delayed-ACK to add ~40 ms of stall per command. With clients issuing `GETATTRS` / `GETHIST` / `GETLOG` sequentially per case, that compounded to seconds per full fetch (the production symptom: ~7 s to load 67 open cases in `curitz`).

This PR coalesces each command's response into a single `transport.write()` — emulating what the legacy Tcl server does via `fconfigure -buffering line -blocking false`. No protocol or wire-format change; existing clients are unmodified.

See #543 for the full diagnosis. The measurement harness referenced from the issue (a public gist) reproduces the before/after numbers locally without needing root or `tcpdump`.

### Implementation notes

`_respond_raw` now buffers each line into a per-protocol `bytearray`, and `_flush_response` writes the whole buffer in a single `transport.write` call. Flush is invoked at exactly four lifecycle points, each a distinct moment when a response is complete:

* after the greeting in `connection_made`
* in `_dispatch_command`, for synchronous paths that returned without scheduling a task (unknown command, unauthenticated, argument errors)
* in `_run_async_responder`'s `finally`, covering every async command
* before `transport.close()` in `do_quit`

The existing dispatch body has been extracted verbatim into a private `_dispatch_to_responder` helper so the wrapper can flush on the sync return path without indenting the dispatch logic.

### Headline numbers (scenario B from the gist — 100-event fixture, client through a zero-delay local TCP proxy)

|                          | Pre-fix | Post-fix |
|--------------------------|---------|----------|
| Total fetch              | 7.07 s  | ~0.05 s  |
| `GETATTRS` mean          | 41 ms   | < 1 ms   |
| Server→client TCP chunks | ~1220   | 304      |

### This pull request
* ~~adds/changes/removes a dependency~~
* ~~changes the API~~ — only the internal write pattern; the line-oriented Zino-1 wire protocol is byte-for-byte identical.

## Contributor Checklist

* [x] The first line of the commit message continues the sentence "If applied, this commit will ...", starts with a capital letter, does not end with punctuation and is 50 characters or less long.

### DIFF
--- a/src/zino/api/legacy.py
+++ b/src/zino/api/legacy.py
@@ -84,6 +84,7 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
         self._multiline_buffer: List[str] = []
         self._authentication_challenge: Optional[str] = None
         self._responders = self._get_all_responders()
+        self._output_buffer = bytearray()
 
         self._state = state if state is not None else ZinoState()
         self._config = config if config else zino.state.config
@@ -114,6 +115,8 @@ class Zino1BaseServerProtocol(asyncio.Protocol):
             self.server.active_clients.add(self)
         self._authentication_challenge = auth.get_challenge()
         self._respond_ok(f"{self._authentication_challenge} Hello, there")
+        # No dispatcher wraps this, so flush the greeting explicitly.
+        self._flush_response()
 
     def connection_lost(self, exc: Optional[Exception]) -> None:
         _logger.info("Client disconnected: %s", self.peer_name)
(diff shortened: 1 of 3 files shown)


src/zino/api/legacy.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         self._multiline_buffer: List[str] = []
         self._authentication_challenge: Optional[str] = None
         self._responders = self._get_all_responders()
+        self._response_buffer = bytearray()
```
lunkwill42: Potentially rename to _output_buffer to be consistent with _input_buffer

lunkwill42: This was a comment from an IRL code review with the team, and although the semantics are slightly different between the buffers, the naming symmetry is a good argument. Force-pushed updates.

aleksfl (approved):
Code looks good to me. Considered some comments on the code comments, but concluded that they are decently descriptive.

PR #548, opened 2026-06-03
Always emit `port` attribute on portstate events

lunkwill42:
## Scope and purpose

Fixes #546.

Single-point fix to [`Event.model_dump_simple_attrs()`](https://github.com/Uninett/zino/blob/3d19876/src/zino/statemodels.py#L307) (the serializer the legacy Zino-1 protocol API uses to dump event attributes) so that fields with a non-`None` declared default (e.g. `port: Optional[str] = ""`) fall back to that default on the wire instead of being silently dropped when their in-memory value happens to be `None`.  Restores the implicit contract that legacy clients (`ritz`, `curitz`, anything `zinolib`-based) rely on.

The complementary question — *why* `event.port` can become `None` in the first place (most likely a link trap arriving for an ifIndex Zino 2 hasn't polled yet) — is tracked as a separate feature request in #547.

## Contributor Checklist

* [ ] ~~Added/changed documentation~~ — bug fix with no observable change beyond what the changelog fragment describes
* [x] The first line of the commit message continues the sentence "If applied, this commit will ...", starts with a capital letter, does not end with punctuation and is 50 characters or less long.  See https://cbea.ms/git-commit/
* [x] If applicable: Created new issues if this PR does not fix the issue completely/there is further work to be done — follow-up filed as #547

### DIFF
--- a/changelog.d/546.fixed.md
+++ b/changelog.d/546.fixed.md
@@ -1,0 +1,1 @@
+Fix legacy Zino-1 protocol clients (`ritz`, `curitz`, `zinolib`-based) crashing when a portstate event lacks a `port` attribute on the wire.  The API now always emits the attribute, falling back to an empty value when needed.
--- a/src/zino/statemodels.py
+++ b/src/zino/statemodels.py
@@ -2,6 +2,7 @@
 
 import datetime
 import fnmatch
+import functools
 import logging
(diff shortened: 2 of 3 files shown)


stveit (approved):
non none none non none non none none none non non none

PR #557, opened 2026-06-26
Fix Docker image build triggered by GitHub releases

lunkwill42:
## Scope and purpose

Publishing a GitHub release has never produced a Docker image. The release-triggered build passed the full git ref (`refs/tags/X`) to `docker/metadata-action`'s `type=semver` tag, which only accepts a bare version — so it generated no tags and `build-push-action` aborted with *"tag is needed when pushing to registry"*. Only the manual `workflow_dispatch` path worked, because it derived a clean tag via `gh release view`.

This emits the clean release tag (`github.event.release.tag_name`) on the release path, and adds an optional `version` input to the manual trigger so a specific past release can be rebuilt without cutting a new one (defaulting to the latest release).

A `release`-triggered run uses the workflow file as it exists at the release's tagged commit, so this fix only takes effect for releases cut after it merges; existing releases can instead be rebuilt on demand with the new `version` dispatch input.

User-controlled values (the release tag and the dispatch `version` input) are passed via environment variables rather than interpolated into the shell, to avoid script injection near the `packages: write` token.

### This pull request
* changes CI configuration only

### Testing

It's always finicky to "test" a github workflow. I already had a Zino fork, so I tested this workflow there and produced this package: https://github.com/lunkwill42/zino/pkgs/container/zino

## Contributor Checklist

* [ ] ~~Added/amended tests for new/changed code~~ (CI-only change)
* [x] Linted/formatted the code (pre-commit hooks pass; no Python changed)

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -20,22 +20,23 @@ jobs:
       packages: write
 
     steps:
-      - name: Determine ref to build
-        id: ref
+      - name: Determine release tag to build
+        id: release
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
-          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
-            tag=$(gh release view --repo "${{ github.repository }}" --json tagName -q .tagName)
-            echo "ref=$tag" >> "$GITHUB_OUTPUT"
+          if [ "${{ github.event_name }}" = "release" ]; then
+            tag="$RELEASE_TAG"
           else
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-images.yml
```diff
@@ -20,22 +20,23 @@ jobs:
       packages: write
 
     steps:
-      - name: Determine ref to build
-        id: ref
+      - name: Determine release tag to build
+        id: release
```
johannaengland: Same comment as in Argus: This id could be more identifying to what is happening

johannaengland (approved):

.github/workflows/docker-images.yml
lunkwill42: same resolution as in argus then :)

PR #558, opened 2026-06-26
Add the Zino logo and update the docs copyright year

lunkwill42:
## Scope and purpose

The published documentation (Read the Docs) carried no branding — the Alabaster sidebar showed a plain "Zino" text title. This adds the Zino wordmark logo to the docs sidebar so the published docs look branded and recognisable.

The black-background logo variant is used because it reads correctly against Alabaster's white sidebar; the transparent/white variant (white text) would be invisible there. The logo is wired through Alabaster's own `logo` theme option rather than Sphinx's `html_logo`, because this theme only honours the former — a [gotcha now recorded in a comment beside the config](https://github.com/Uninett/zino/blob/0f6bab83771dbecf5a0802b25b23805f222b86c8/docs/conf.py#L43-L48) so it survives a future theme or config change.

While here, the docs copyright end-year is derived at build time instead of being hard-coded (separate commit), so it no longer needs a manual bump each year. Read the Docs builds stay reproducible because Sphinx rewrites the year from `SOURCE_DATE_EPOCH` (the last-commit date) when that is set.

No dependency or API changes.

## Contributor Checklist

* [ ] ~~Added a changelog fragment for [towncrier](https://github.com/Uninett/zino/blob/master/README.md#before-merging-a-pull-request)~~ — docs/branding-only change, no user-facing software effect.
* [ ] ~~Added/amended tests for new/changed code~~ — not applicable to a logo asset + Sphinx config change.
* [ ] ~~If applicable: Created new issues if this PR does not fix the issue completely/there is further work to be done~~ — not applicable.


aleksfl (approved):
Looks good, works for me!

docs/conf.py
```diff
@@ -5,11 +5,12 @@
 
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+import datetime
 import os.path
 import sys
 
 project = 'Zino'
-copyright = '2024-2025, Sikt - The Norwegian Agency for Shared Services in Education and Research'
+copyright = f'2024-{datetime.date.today().year}, Sikt - The Norwegian Agency for Shared Services in Education and Research'
```
johannaengland, line 13:
```suggestion
copyright = '2024-%Y, Sikt - The Norwegian Agency for Shared Services in Education and Research'
```

Ref: https://www.sphinx-doc.org/en/master/usage/configuration.html#config-copyright

Sphinx needs to be updated to 8.1 for it to work

lunkwill42: nice :) but we would have to drop support for python 3.9 first, making this a follow-up issue.  Can you post one, @johannaengland ?

johannaengland: Can do!

PR #560, opened 2026-06-26
Fix SNMP trap reception when running Zino in Docker

lunkwill42:
## Scope and purpose

When Zino runs in Docker with the bundled compose setup, it silently receives no SNMP traps. Zino matches each trap to a device by its source IP address, but Docker's default bridge networking with published ports rewrites the source of incoming UDP packets to the bridge gateway — so every trap is unmatched and dropped. This has bitten a real deployment (Zino under Docker on a VM): polling works, traps never register.

The compose setup now uses host networking, which preserves the real source address. Binding the privileged trap port 162 requires starting as root, so the container starts as root and then drops privileges — to a built-in unprivileged `zino` user by default (never staying root), or to your own `HOST_UID:HOST_GID` so bind-mounted config and state files stay owned by you. Dropping to a host UID required teaching Zino's `--user`/`process.user` to accept a numeric `UID` or `UID:GID` (previously name-only via `getpwnam`, which can't target a UID with no `/etc/passwd` entry in the container).

The docs gain a "why host networking is required" section and note the external `nmtrapd` multiplexer as the alternative where host networking isn't possible.

### This pull request
* changes the Docker image, its compose setup, and its documentation
* adds a numeric `UID`/`UID:GID` form to the privilege-drop option

## Contributor Checklist

### DIFF
--- a/changelog.d/+privilege-drop-numeric-uid.added.md
+++ b/changelog.d/+privilege-drop-numeric-uid.added.md
@@ -1,1 +1,1 @@
-`--user` and the `process.user` setting now accept a numeric `UID` or `UID:GID` in addition to a user name, so Zino can drop privileges to a UID that has no entry in `/etc/passwd` (e.g. inside a container)
+`--user` and the `process.user` setting now accept a numeric `UID` or `UID:GID` or, as previously, a user name, so Zino can drop privileges to a UID that has no entry in `/etc/passwd` (e.g. inside a container)
--- a/changelog.d/+privilege-drop-numeric-uid.added.md
+++ b/changelog.d/+privilege-drop-numeric-uid.added.md
@@ -1,0 +1,1 @@
+`--user` and the `process.user` setting now accept a numeric `UID` or `UID:GID` in addition to a user name, so Zino can drop privileges to a UID that has no entry in `/etc/passwd` (e.g. inside a container)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,17 +6,20 @@ services:
     image: zino:latest
     # Option 2: Fetch image externally
     #image: ghcr.io/uninett/zino:latest
+    # Host networking is required so Zino sees the real source address of
+    # incoming SNMP traps. With bridge networking and published ports, Docker
+    # rewrites the source to the gateway address and Zino ignores the traps,
+    # because it matches them to devices by source address. See
+    # docs/howtos/use-docker-image.rst for the full explanation.
+    network_mode: host
     volumes:
-      # Mount the current directory into /zino inside the container
-      # It needs to contain the various necessary configuration files
+      # Mount the current directory into /zino inside the container.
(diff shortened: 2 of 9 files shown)


docker-compose.yml
```diff
@@ -6,17 +6,20 @@ services:
...
-    ports:
-      # Default trap port
-      - "162:162/udp"
-      # SNMP "ping" agent, for clients that want to verify Zino uptime
-      - "8000:8000/udp"
-      # Map the API and notification ports
-      - "8001:8001"
-      - "8002:8002"
-    # Example custom trap port configuration, must match with port mapping
-    #command: "--trap-port 1162"
```
hmpf, line 22: So all of this is unneccessary when using `network_mode: host`?

hmpf (commented):
Where/how's the trap port set? It's not explicit in the compose-file anymore.

changelog.d/+privilege-drop-numeric-uid.added.md
```diff
@@ -0,0 +1 @@
+`--user` and the `process.user` setting now accept a numeric `UID` or `UID:GID` in addition to a user name, so Zino can drop privileges to a UID that has no entry in `/etc/passwd` (e.g. inside a container)
```
johannaengland: Difficult to phrase but to me this first sounded like you can give it an UID and a user name 
```suggestion
`--user` and the `process.user` setting now accept a numeric `UID` or `UID:GID` or, as previously, a user name, so Zino can drop privileges to a UID that has no entry in `/etc/passwd` (e.g. inside a container)
```

johannaengland (approved):
Just a nitpick about a changelog entry

johannaengland:
> Where/how's the trap port set? It's not explicit in the compose-file anymore.

The default value of the parameter `trap-port` of the `zino` command is 162 - https://github.com/Uninett/zino/blob/4757a78a0602230a43f9ef4293de004fd6dc4201/src/zino/zino.py#L463

docker-compose.yml
lunkwill42: Basically, yes. All that port-faffing is necessary when running in a bridged network, to forward ports from the host to the container inside the bridged network.  

Running in host network mode makes the docker container attached to the host machine's network stack, rather than a bridged one (i.e. there's no isolation of the network namespace, just the process space).

lunkwill42:
@hmpf said:
> Where/how's the trap port set? It's not explicit in the compose-file anymore.

It was never explicit, it was just a comment with an example. If you, for some reason, want to change it away from the standard 162 that all agents expect (like during development), the docs do provide info on how to change it using `--trap-port`, so I basically didn't care to keep the comment.

changelog.d/+privilege-drop-numeric-uid.added.md
lunkwill42: This is probably clearer, adopted, thanks :)

